PR #8, opened 2011-03-23
Shared sockets

gnosek:
As a follow up to the way overdue shared sockets patch, here it is again, refactored and (IMHO) cleaner than the previous approaches.


adamn:
Just guessing ... but it seems like after 5 years there's no hope for this PR and it should be closed.

PR #84, opened 2012-01-31
Fixing %(here)s in included configs, fixes #81

### LINKED ISSUE #81: Make a counterpart of %(here)s for included configs
Hi,

The `%(here)s` placeholder is very useful to set up processes relative to the config file path. However, `%(here)s` used in included configs resolve to the path of the main config, not the included one.

I suggest another placeholder called, let's say `%(include_here)s`, which would resolve to the path of the included config that contains it.

It would be very useful for the following scenario. I have a webapp which runs some background processes (consuming messages), there is a supervisord config in the repository of the webapp, that is included by the main config, so my `/etc/supervisord.conf` looks like:

    ; ... blabla
    [include]
    files = /var/www/config/supervisord_processes.ini

The `supervisord_processes.ini` contains the description of the processes of the webapp, so whenever I do a deployment, I just send a HUP to supervisord, reload the config and I have my background processes restarted with the new code.

The problem is that in the included config I have to use absolute path to my background processes. This is a little bit chaotic when working with several environments (devel, staging, production...) where the place where my code resides could be different.

This is why I need the `%(include_here)s`, to be able to refer to my processes relatively to the checkout path.

By the way, let me tell, I'm amazed by this project, it's really great and exactly was I was looking for.

Thank you for the help!

tcz:
As we talked, we need to problem with %(here) to be fixed.
I kept my own hacky approach and fixed it without breaking unit tests.
However, maybe adding some tests for this feature would be nice.

Let me know if you can pull this.

Cheers

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -509,7 +509,21 @@ class ServerOptions(Options):
                     self.parse_warnings.append(
                         'Included extra file "%s" during parsing' % filename)
                     try:
-                        parser.read(filename)
+                        # copying all values from included config to this config
+                        # this way we can add extra values
+                        include_parser = ConfigParser.RawConfigParser()
+                        temp_include_file = tempfile.TemporaryFile();
+                        include_parser.read(filename)
+
+                        for include_section in include_parser.sections():
+                            for include_section_key, include_section_value in include_parser.items(include_section):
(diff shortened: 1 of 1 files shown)


mnaberez:
There is a new pull request in #567 that does this without the issues noted above, and the original issue in #81 is still open.  I am going to close this pull request and we can continue discussion on those.  Thank you for your work on this.

PR #135, opened 2012-07-18
this adds support for status output for groups

simonjj:
example usage: 

`supervisorctl status backend:*`

example output:

```
backend:backend_0                RUNNING    pid 3830, uptime 0:00:07
backend:backend_1                RUNNING    pid 3831, uptime 0:00:07
```


werdender:
supervisorctl status | grep <group name>

PR #162, opened 2012-11-13
Let the supervisord daemon log to syslog.

fcuny:
When the options `stdout_logfile` or `stderr_logfile` are set to
"syslog", the children are logging to the syslog service,
but the daemon is still logging to a file.

By letting the possibility to set the value for `logfile` to
"syslog", we can have the same behavior for the daemon.

### DIFF
--- a/docs/logging.rst
+++ b/docs/logging.rst
@@ -14,7 +14,11 @@ messages about its own health, its subprocess' state changes, any
 messages that result from events, and debug and informational
 messages.  The path to the activity log is configured via the
 ``logfile`` parameter in the ``[supervisord]`` section of the
-configuration file, defaulting to :file:`$CWD/supervisord.log`.
+configuration file, defaulting to :file:`$CWD/supervisord.log`.  If
+the value of this option is the special string "syslog", the logs of
+the :program"`supervisord` will be routed to the syslog service
+instead of being written to a files.  As a consequence, the options
+``logfile_maxbytes`` and ``logfile_backups`` will be ignored.
 Sample activity log traffic is shown in the example below.  Some lines
 have been broken to better fit the screen.
 
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -447,7 +447,11 @@ class ServerOptions(Options):
         else:
             logfile = section.logfile
 
-        self.logfile = normalize_path(logfile)
+        if logfile != 'syslog':
+            # if the value for logfile is "syslog", we don't want to
+            # normalize the path to something like $CWD/syslog.log, but
+            # instead use the syslog service.
+            self.logfile = normalize_path(logfile)
 
         if self.pidfile:
             pidfile = self.pidfile


Siecje:
I'm curious why this has not been accepted. It is a minor change and a desired feature.

mnaberez:
@Siecje There are a large number of patches tagged [syslog](https://github.com/Supervisor/supervisor/labels/syslog) with overlap between them.  They need to be reviewed together.

Siecje:
@mnaberez There doesn't seem to be another pull request with this feature. This has sat for 3 years and the functionality is already there for programs just not supervisord itself.

mnaberez:
@Siecje There's an alternate one [right above](https://github.com/Supervisor/supervisor/pull/162#issuecomment-17540811) and I've seen this in at least one other patch if not more.  The syslog patches probably need to be reviewed together.

samv:
Sure, here's your summary.  I'm including it on this one because it's the oldest with that tag, and this is where the discussion is happening.

This one: #162 - created by an ex-colleague of mine, @franckcuny - is where I started with my own PR.
The included patch above as well as #286, #334 and #563 I believe all deliver features included in my PR: mostly relating to configuring the syslog tag (application name), as well as syslog facility and priority.

#446 is genuinely different and relates to remote syslog logging.  This is especially important for docker users who are using supervisor as the top-level process in a container, because they don't have a 'syslog' daemon running locally.  If you merge this PR, I'll submit a new one which cleans this up and allows it.  Personally I think to be consistent you should either support syslog properly in core, or clearly deprecate it and instead point people to the plug-in.

mnaberez:
> This one: #162 - created by an ex-colleague of mine, @franckcuny - is where I started with my own PR.  The included patch above as well as #286, #334 and #563 I believe all deliver features included in my PR: mostly relating to configuring the syslog tag (application name), as well as syslog facility and priority.

Note for future review: "my own PR" that @samv references above is #289.

jaraco:
Also, don't forget #241, which supersedes #162 and was accepted and sits in master but has never been released (and whose configuration options now suggest that 4.0.0 will include the changes). #241 addresses the shortcoming of #162 indicated in #182.

jaraco:
My recommendation - close #162, (optionally) release 4.0.0 with the changes in master, accept #289 which fixes #286, #334, and #563, and release that as a backward-incompatible release (4.0 or 5.0). Address #446 subsequently.

The important thing here is to make releases on which people may rely and iterate and refine. Instead, we have lots of private forks that organizations are running because the functionality has not been released.

mnaberez:
>  (optionally) release 4.0.0 with the changes in master,

It is unfortunate that the syslog changes are tied up with the changes on master.  Unfortunately, the master branch has Python 3 support which is very broken (see issues tagged "python3" and readme) and the people who were working on that have stopped.  An alternative might be to make a new 3.x release with the syslog changes.

jaraco:
@mnaberez: A quick browse through those issues, some of them are specific to Python 3 (which is currently unsupported, so a release note stating that Python 3 is still not perfect should suffice), but others seem to be regressions brought about by the attempt at Python 3 support. I suggest we bisect those two failure modes and address the latter, bringing the master back to a stable state independent of supporting Python 3. I'd flag those issues as 'blocker' or similar.

mnaberez:
> so a release note stating that Python 3 is still not perfect should suffice

I'm not willing to release software that I know will crash hard with something as simple as outputting a byte that is not valid UTF-8.  Companies run this software in production and that's far from an edge case.

Siecje:
Merging this pull request is backward compatible.

jaraco:
> That's far from an edge case. Companies run this software in production and don't always read all the details in the release notes. 

Fair enough. So hard fail when `sys.version_info > (3,)` but let's not let the Python 3 support be an unnecessarily high barrier to stability of master.

jerryhebert:
Hey folks, is there any update on this and all of the related syslog PRs? I'm trying to understand the best approach to output to syslog with my own preferred local facility but that functionality seems to have been locked up in one PR or another for several years now. I am not sure how to proceed, but it certainly seems like making yet another PR to add this functionality would not help matters.

Any advice would be appreciated, thanks!

xiva-wgt:
+1

xiva-wgt:
Docker container set log to nsyslog on host. Only supervisord write log files into container.
Need log to syslog without filehandler.

vlad-tuznichenko:
+1

PR #215, opened 2013-02-08
Add very basic Linux control groups support

adembo:
Control groups (cgroups) are a Linux kernel feature that can be used for
resource management and isolation between processes. In my case, I have an
out-of-band application that manages the cgroups themselves, but as I use
supervisor to manage process lifecycle, it's well positioned to attach
processes to cgroups, and to reattach them if they're autorestarted.

Here we do that, with two main changes:
1. After forking a process, supervisor will attach it to every desired
   cgroup by writing its pid to each 'tasks' pseudofile. Note that cgroup
   membership is per-thread (not per-process), but at this stage in the
   process' life, it should be single threaded.
2. When queried for process information, supervisor will also mention
   if the process has cgroups or not. Ideally we'd take into account whether
   attaching actually succeeded, but effecting that change in the parent is
   tough from the forked child.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -231,7 +231,8 @@ Process Control
              'exitstatus':     0,
              'stdout_logfile': '/path/to/stdout-log',
              'stderr_logfile': '/path/to/stderr-log',
-             'pid':            1}
+             'pid':            1,
+             'has_cgroups'     0}
 
         .. describe:: name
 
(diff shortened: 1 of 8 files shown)


gtt116:
I love the idea to make supervisor support controlling process's cgroup. As we know [systemd support cgroup now](https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html), if supervisor can support this feature will be amazing.

luto:
@mnaberez is there any interest in this feature? I'd love to code something similar to this up for a PR.

pavel-orekhov:
> @mnaberez is there any interest in this feature? I'd love to code something similar to this up for a PR.
Yes. I have interest is this feature.

JoveYu:
may be pid should append to task not overwrite

f18m:
Hi all,
I'd love to have such feature built into supervisord as well... any comment from SupervisorD maintainers on this PR ?

Thanks

PR #229, opened 2013-04-11
add rlimit configuration for programs

davidbirdsong:
Hi, I've used a hacked up supervisord for years with the options:

``` ini
[program:prog1]
command=/bin/cat
minprocs=4096
minfds=4096
stacksize=10240
```

Which directs supervisor to call setrlimit to the values just after forking to launch the program. For RLIMITS with a soft and hard limit, I've just forced the 'min' value to both since most programs under supervisord are not equipped to call setrlimit on their own behalf.

I think this is generally useful, and I'd love to stop having to manually patch to get updates. 

I ran all tests and had to run tests with root privileges to get the final tests to pas.

Let me know if I've abused any code or if any code style is not good.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1255,6 +1255,19 @@ class ServerOptions(Options):
                 'name':'RLIMIT_NPROC',
                 })
 
+        if hasattr(resource, 'RLIMIT_STACK'):
+            limits.append(
+                {
+                'msg':('The maximum stack size in bytes'
+                       'to run this program is %(min)s as per the "stacksize" '
+                       'command-line argument or config file setting. '
+                       'The current environment will only allow you '
+                       'to allocate %(hard)s bytes for stack.'),
+                'min':self.config.stacksize,
+                'resource':resource.RLIMIT_STACK,
+                'name':'RLIMIT_STACK',
+                })
+
         msgs = []
 
         for limit in limits:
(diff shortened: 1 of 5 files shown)


ayashjorden:
Hi all,
Any update when this PR is going to be merged?
Setting the nofiles limit is really important for high-throughput processes.

Thanks,
Yarden

Fizzadar:
I'm assuming this is never going to be merged? Would be nice to know for sure. Moving to either circus or runit in the meantime...

mpictor:
If I understand correctly, the implementation of supervisor is such that a process can be started as a non-root user, **with more privileges than that user is allowed** - bypassing pam, limits.conf, etc. If so, this is an _insane_ design and I'll be advising my company to run the hell away.

Even if this particular problem is fixed, I'll have second thoughts about using supervisor: developers who do not understand security or don't take it seriously are likely to continue creating security problems.

woosley:
I need this feature badly also, maybe time to take a look at circus?

antgel:
We moved to circus, which has it's own "idiosyncrasies". Now looking at systemd. :)

carun:
After 4 years, will this PR be merged?

sar009:
WTF are people at supervisor even listening. It is end of 2019 and still no resolution for a critical request.

davidbirdsong:
> After 4 years, will this PR be merged?

I opened this PR 6 1/2 years ago, but at this point let's all stop bagging on this project and complaining about this PR.  Many suitable alternative projects with active owners have emerged since.

PR #241, opened 2013-05-25
Separate log targets for syslog (and other streams)

jaraco:
My last pull request added syslog support by adapting a patch supplied by a user to allow specifying 'syslog' as a special "filename" for the log targets. If specified, the setting would then direct the stream for that channel (stdout or stderr) to syslog.

Unfortunately, by invoking this functionality, Supervisord no longer can log to a file nor can it provide a copy of the log in the API, the latter of which was very important for our application.

This new patch, a series of changes that I worked over time and which we're now running in our production environment, allows the user to additionally configure two new settings, 'syslog_stdout' and 'syslog_stderr'. If either are indicated and True, that channel will be directed to syslog, independent of the setting for the file stream.

This change required substantial refactoring to implement as an additional, optional, stream, and I have taken care to describe those changes in deliberately small steps for easier review. All of the changes in this pull request were made as a bare minimum approach to achieve the stated goal in a robust way and no more.

Again, with tests passing and with this running in our production environment, I have confidence that it is a solid improvement. I hope you find the same. Please consider it for inclusion upstream.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1604,11 +1604,11 @@ class ProcessConfig(Config):
         'name', 'uid', 'command', 'directory', 'umask', 'priority',
         'autostart', 'autorestart', 'startsecs', 'startretries',
         'stdout_logfile', 'stdout_capture_maxbytes',
-        'stdout_events_enabled',
+        'stdout_events_enabled', 'stdout_syslog',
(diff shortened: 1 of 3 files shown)


jaraco:
> adapting a patch supplied by a user

When I mentioned that, I meant the patch proposed in #162.

jaraco:
@mnaberez Can you tag this issue with syslog?

PR #274, opened 2013-08-06
Switching docstrings to sphinx-style

EvaSDK:
The output of the sphinx generated documentation is not as nice as it could be.

The patch adds sphinx-style docstring support to gettags function so that everything should continue to work and have the benefit of a nicely integrated documentation.

I did not drop the java/epydoc style support since it probably needs a deprecation period for third parties, etc.


mnaberez:
There are so many plugins now that I would rather not change this, sorry.

Issue #116, opened 2012-05-09
supervisorctl pid versus return codes


ctheune:
`supervisorctl pid` can be used to
- retrieve the pid of (the running) supervisor daemon itself
- pid(s) of the child processes

I can't see what the intention is about this command using specific return codes. I see that it can sometimes propagate an exception causing Python to set a non-zero exitcode.

OTOH if supervisor doesn't run it simply prints a message (that is not a pid) and keeps a zero exit code.

We're writing some code that needs to figure out whether supervisor is running and we find ourselves to either try converting the output of `supervisorctl pid` to int (and start if that fails) or calling `supervisorctl reload` sometimes (which returns a non-zero exit code if it doesn't run yet).

We like to see `supervisorctl pid` to help us out with reasonable exit codes. I'm happy to contribute code if I know what your intentions are.

Issue #24, opened 2011-06-26
exit code allway zero


ramary:
I want to use the exit code for nagios but it's allways 0.

ecample 1:
zope@plone:~> ./bin/supervisorctl status
http://127.0.0.1:9001 refused connection
zope@plone:~> echo $?
0

Could you change exit LSB conform ?
http://refspecs.linuxfoundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/iniscrptact.html

example 2:
./bin/supervisorctl status
balancer                         RUNNING    pid 23753, uptime 0:00:50
instance0                        RUNNING    pid 23751, uptime 0:00:50
instance1                        EXITED     Jun 26 05:40 PM
varnish                          RUNNING    pid 23754, uptime 0:00:50
zeo                              RUNNING    pid 23750, uptime 0:00:50
zope@plone:~> echo $?
0

exi:
This ticket is years old and concerns a serious shortcoming. Could we bump it's priority or provide some kind of sensible workaround?

gdm85:
@exi I have been using @lukeweber's patch, but obviously already switching to [runit](http://smarden.org/runit/) for the long term/production.

rtwair:
Labeling this as an enhancement makes absolutely no sense.

+1 - definitely a bug

neutrinus:
+1

This issue makes us cry at least once a week. As supervisord is not returning anything but 0, ansible has to parse the "human" strings to check the state.

mattwiller:
+1 — this is a serious issue that makes it very difficult to integrate supervisor with our deployment infrastructure

castarco:
+1 - That's a serious bug. This shoudn't be seen as a low priority enhancement

It difficults the integration of supervisord with other automatization tools (rather than checking the return code, the only generic way to know if the command failed is to analyze the supervisorctl output to stdout....).

vvanscherpenseel:
+1, this is a bug

lukeweber:
+1 for getting this fixed. It's pretty unusable for production for me without the patch. I've been using my patched version for a year now in production. @mnaberez would you merge this if there were tests or is there some other holdup? At 4 years pending I'm a bit curious why it's still not done. I recall it looking like a bit of a rewrite would be in order to test exit codes adequately.

mnaberez:
>  @mnaberez would you merge this if there were tests or is there some other holdup?

[...]

PR #292, opened 2013-10-01
Add exit codes to supervisorctl

GotenXiao:
This is a first pass at addressing #24 and #116, and allows regular automation of starting supervisor if it is not already running (e.g. supervisorctl status &>/dev/null || supervisord).

### DIFF
--- a/supervisor/supervisorctl.py
+++ b/supervisor/supervisorctl.py
@@ -86,7 +86,7 @@ class fgthread(threading.Thread):
     def localtrace(self, frame, why, arg):
         if self.killed:
             if why == 'line':
-                raise SystemExit()
+                raise SystemExit(1)
         return self.localtrace
 
     def kill(self):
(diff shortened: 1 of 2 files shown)


IvanAnishchuk:
I think exit codes on `supervisorctl status` differing depending on the status (e.g. 0 for running, 1 for stopped, 2 for error, etc.) could also be helpful. But we need to be careful and separate codes from socket communication errors and other supervisor's internal issues (positive/negative codes?)

gdm85:
@IvanAnishchuk in my opinion the exit code shall communicate the result of the command that user asked to execute.

If supervisorctl can read and report successfully the result of ``status``, then there is no need to abuse the exit code with a different functionality than most command line interface utilities do.

Same way if ``start`` or ``stop`` failed, the exit code should tell about it. I am currently using this PR for this reason.

msabramo:
This no longer merges cleanly.

mnaberez:
Thanks for your work on this patch.  There's now a newer patch in #668 that should handle these conditions and adds exit codes for other error conditions as well.  I intend to merge #668, so I'm going to close this PR in favor of that one.  I'd appreciate if you could please check out that PR and let us know if it will work for you.  I'll credit you in the changelog for helping add exit codes to `supervisorctl`.

PR #296, opened 2013-10-10
dont show ERROR when started already

steverweber:
dont show ERROR when started already


mnaberez:
Closing since this fails the unit tests.  There are a number of other tickets tagged `supervisorctl`, some of which have overlap with this one.

PR #297, opened 2013-10-17
look for SUPERVISOR_SERVER_URL when connecting with supervisorctl

plockaby:


mnaberez:
I have decided not to merge this, sorry.  We don't have any other individual config file options that are overridden by environment variables.  I don't think we should add an option that would allow the URL to be set but not the other options, as many/most users run it with authentication enabled.  The shell alias is probably the best solution, or perhaps we can add another way to look for the supervisord.conf file.

PR #301, opened 2013-10-21
Print a warning if the included config doesn't contain at least one valid section

Kami:
This pull request adds support for printing a warning if the included config file doesn't contain at least one valid section.

This should make spotting issues which are related to typos a lot easier. For example, today I just spent a bunch of time debugging an issue which was related to me misspelling "program" as "progam" (yes, I do use spell checking in my text editor, but it's not so smart to recognize ":" as a delimiter).

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -56,6 +56,14 @@ mydir = os.path.abspath(os.path.dirname(__file__))
 version_txt = os.path.join(mydir, 'version.txt')
 VERSION = open(version_txt).read().strip()
 
+# A list of valid sections in the config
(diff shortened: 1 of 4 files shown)


mnaberez:
I appreciate the work that went into this, but I have decided to merge it, sorry.  I'd like to improve error detection in the config file in general, but change adds a lot of code to detect this one condition.  It also requires us to start maintaining the `VALID_` lists, which is error prone.  There is actually one such error in it already (`'eventlistener'` not `'eventlistener:'`).  I'd be open to reviewing another pull request if there was a simpler way to detect this condition.  I will also try to keep this in mind for future work on the config file parsing.

PR #329, opened 2013-12-17
Parse env configuration in posix mode

### LINKED ISSUE #328: Not possible to have quotes in environment variable settings
It is not possible to pass environment variable values containing quotes, e.g.

```
[program:x]
command=bash -c export
environment=MYSQL_PW="5ny6{wo'\"7n"
```

throws

```
ERROR: CANT_REREAD: No closing quotation
```

Currently shlex is used in legacy mode to parse the env settings, which does not allow any kind of escaping and also has some other odd edge cases (http://docs.python.org/2/library/shlex.html#parsing-rules). I'd propose switching the lexer to posix mode. Ideally, we would like to be able to dump any string dictionary and have it load correctly as environment for the process.

sfriesel:
Fixes #328

This allows (escaped) quotes in the values as well as empty
values.

### DIFF
--- a/supervisor/datatypes.py
+++ b/supervisor/datatypes.py
@@ -81,7 +81,7 @@ def dict_of_key_value_pairs(arg):
     """ parse KEY=val,KEY2=val2 into {'KEY':'val', 'KEY2':'val2'}
         Quotes can be used to allow commas in the value
     """
-    lexer = shlex.shlex(arg)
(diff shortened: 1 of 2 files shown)


dgouldin:
It looks like `posix` was added in 2.3, unless I'm misreading: https://docs.python.org/release/2.4.4/lib/module-shlex.html

Is that not correct?

PR #330, opened 2013-12-18
Store raw parameters from .conf files in process/group config objects

janitha:
Store raw parameters from the .conf file for each process/group with the config objects. This allows setting custom parameters per program/group that can be used by by plugins utilizing the rpc interface.

Example:
[program:xyz]
command=/bin/cat
custom_param1=12345
custom_foo=bar

All key/value pairs in this section would be available via this process's .config.raw_params as a dict

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -628,7 +628,8 @@ class ServerOptions(Options):
                 group_processes.extend(processes)
             raw_params = dict(parser.items(section))
             groups.append(
-                ProcessGroupConfig(self, group_name, priority, group_processes, raw_params)
+                ProcessGroupConfig(self, group_name, priority, group_processes,
+                                   raw_params=raw_params)
                 )
 
         # process "normal" homogeneous groups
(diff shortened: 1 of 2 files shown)


mnaberez:
Thanks for your work on this but I have decided not to apply it at this time.  I think this is a neat idea but it has the unfortunate side effect of making common config file errors undetectable.  

An issue that has come up in other tickets is that users sometimes misspell option names (e.g. `log_level` instead of `loglevel`).  `supervisord` currently ignores these when processing the config file and this can lead to unpredictable behavior.  It would be much better if we recognized bad names and halted immediately with an error.  We may never be able to do that if we apply this patch.

I think it would still be possible to implement this same feature while also detecting config file errors.  Rather than `raw_params` (everything), we could add a mechanism so RPC interface plugins could specify which extra program options they add.

mikeyg123:
We would find this very useful - but take your point about being able to detect config errors.  Perhaps user variables could have a common prefix, e.g. "custom_".
Even better would be to allow these custom parameters to be substituted into the command.

e.g. 
[program:xyz]
custom_file=/dev/random
command=/bin/cat %(custom_file)

PR #333, opened 2013-12-19
Append timestamp to stdout and stderr logs and config flags

nirmalc:
Append timestamp to stdout and stderr logs and config flagsthis uses 'asctime' internal - dict value to append timestamp to stdout , stderr logs.

%s_append_timestamp enables/disables this feature per process.


mnaberez:
I'm going to close this due to activity.

PR #366, opened 2014-01-28
Allow process-level rlimits

DarrenLo:
Supervisor only supports rlimits for supervisor itself. This adds
the ability to add limit_fds, limit_procs, and limit_memlock to the
supervised process' conf file.

Tested that the three options take effect on rhel6.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -805,6 +805,21 @@ class ServerOptions(Options):
             uid = None
         else:
             uid = name_to_uid(user)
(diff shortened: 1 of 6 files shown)


BlaineAtAffirm:
Are either of the rlimits patches going to be added ? Would be very useful to have. Was very confused for a couple days until I found the mail lists about supervisor not working with pam.

raylu:
What is blocking this or the linked PR?

PR #370, opened 2014-02-03
environment_file configuration directive

steinnes:
Due to the fact that we need to have a lot of variables per process
controlled by their respective environment files, we added a new
configuration directive to better facilitate such needs.


mnaberez:
I'm going to close this for the reasons stated above.  If there is demand for this in the future, and [this solution](https://github.com/Supervisor/supervisor/pull/370#issuecomment-33983410) does not work, we can always revisit it.  The [mailing list](https://lists.supervisord.org/mailman/listinfo/supervisor-users) would also be a good place to discuss this.

PR #414, opened 2014-04-12
Added standard Debian location for Supervisor configuration file

kelvinwong-ca:
Debian and Ubuntu packages both place the supervisord.conf file at the following location:
```
 /etc/supervisor/supervisord.conf
```
As an enhancement, could Supervisor attempt to find that conf file in the searchpaths list setting in Options?


IvanAnishchuk:
I agree, shouldn't mess anything up as supervisor already searches in several paths for the config.

mnaberez:
I've left some comments about the search paths over in #576.  I would welcome discussion about this and other search path issues in that thread.

jib:
+1 - we're using the distro packaged version of supervisor right now, and upgrading to latest supervisor stable breaks because this path is not in the search path :(

PR #442, opened 2014-06-10
Make startretries accept -1 in order to never stop retrying

mediocregopher:


kigakree:
+1

mediocregopher:
To the maintainers: This doesn't currently merge but if anyone could give me some indication if this feature is wanted or not I can either fix it up or close the PR. Until then I'm just going to let it sit and not fuss with it.

mnaberez:
I've never had a process running under `supervisord` that took hundreds or thousands of failed respawns and then successfully started, so I don't really understand the use case for this special `-1` value.  It seems like almost the same thing can be achieved by setting `startretries` to a large number.

mediocregopher:
Alright well thanks for looking at the PR anyway :)

gjcarneiro:
FWIW, I often have programs tha

mhf-ir:
+1

triccardi-systran:
This is a useful feature; setting a large number is just a workaround.

Issue #149, opened 2012-09-08
Tail -f option in the web interface doesnt work.


ghost:
On the web page to control the processes, the Tail -f link doesn't work (the server doesn't respond anything, so the browser doesn't show anything). The href attribute of the link is
http://localhost:8080/logtail/recommender_server%3Arecommender_server_01
where recommender_server is the process name. 

Nevertheless the Clear Log link works well. In this case, the href attribute of the link is:
http://localhost:8080/index.html?processname=recommender_server%3Arecommender_server_01&action=clearlog

My operating systems is: OS X Mountain Lion (Mac Book Pro)

; MY CONFIGURATION FILE FOR SUPERVISORD

[unix_http_server]
file=/Users/rmerinogarcia/Proyects/temp/supervisor.sock  
username=user  
password=123     

[inet_http_server]  
port=127.0.0.1:8080  
username=user  
password=123              

[supervisord]
logfile=/Users/rmerinogarcia/Proyects/temp/supervisord.log 
logfile_maxbytes=50MB  
logfile_backups=10  
loglevel=info  
pidfile=/Users/rmerinogarcia/Proyects/temp/supervisord.pid 

[rpcinterface:supervisor]
supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface

[program:recommender_server]
process_name=%(program_name)s_%(process_num)02d
numprocs=1
numprocs_start=1
command=/Users/rmerinogarcia/.Envs_Python_2.7.3/USG_ENGINE/bin/python /Users/rmerinogarcia/.Envs_Python_2.7.3/USG_ENGINE/bin/start_recommender.py -p 91%(process_num)02d
stdout_logfile=/Users/rmerinogarcia/Proyects/temp/recommender.out
redirect_stderr=true
autorestart=true

Shellbye:
I have the same issue but fixed by add childlogdir = %(here)s to [supervisord]. And all my [program] have the directory=%(here)s

ue-it:
It works well with http://127.0.0.1:9001/logtail/test, but not work after nginx with http://my.url/logtail/test   
All the others work well, so I think the problem is not nginx. Can anyone help me to find the reason?

alzmcr:
Same problem here as @ue-it - work well locally but not from remote. Anyone aware of the reason?

sokokaleb:
Same problem. The page says "Error response. Error code 404. Message: Not Found."

samwang1027:
[...]

PR #444, opened 2014-06-11
Try to fix tailf got nothing on Chrome

dawncold:
issue #149 

I have tested Chrome behavior about chunked response. The `Content-Type` should be `text/html` and response content must be a valid html element format, so I wrapped every log row in a `p` element, it is not very beautiful, but it works for me. 

My Chrome Version is 35.0.1916.153 (Official Build 274914) on MacOS.

Here is my test chunked response code, a piece of nodejs code.
```javascript
var http = require('http');

http.createServer(function (request, response) {
    response.writeHead(200, {
        'Content-Type': 'text/html',
        'Transfer-Encoding': 'chunked'
    });
    var x = setInterval(function(){
        response.write('<p>' + new Date() + '</p>' + '\r\n'); 
    }, 1000);
}).listen(8000);
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ htmlcov/
 tmp/
 coverage.xml
 nosetests.xml
+.idea/*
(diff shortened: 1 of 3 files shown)


mnaberez:
Thank you for this pull request but I'm going to close it in favor of #471 or a similar fix because the wrapping breaks `supervisorctl` as noted above.

PR #468, opened 2014-07-21
Use monotonic time for process state tracking

### LINKED ISSUE #281: Crash at  system time change back
There's a reproduce path to make supervisord crash:

 1. set system time 1 year before real time.
 2. killall one_program_under_supervisoring. 
 3. you'll see supervisord crashed.  

Checked the code and see that our process state calculation depends a lot on system time. System time change may cause assertion.

Mook-as:
This makes sure system time changes (e.g. via ntp) doesn't result in odd assertion failures due to processes entering the running state before being started.

Lightly tested on Linux; the BSD code is untested (written based on `man clock_gettime` and FreeBSD headers).  Unfortunately I do not know how to fix it for OSX, and suspect that falling back to `time.time()` is a bad idea.

While running `tox` locally seems to pass, I do not know how much to trust it.

Fixes #281.

### DIFF
--- a/supervisor/compat.py
+++ b/supervisor/compat.py
@@ -151,3 +151,37 @@ try: # pragma: no cover
     import thread
 except ImportError: # pragma: no cover
     import _thread as thread
+
+try: # pragma: no cover
+    from time import monotonic as monotonic_time
+except ImportError: # pragma: no cover
+    if sys.platform.startswith("linux") or sys.platform.contains("bsd"):
+        # Adapted from http://stackoverflow.com/questions/1205722/
+        import ctypes
+        import os
+
+        class timespec(ctypes.Structure):
+            _fields_ = [
+                ('tv_sec', ctypes.c_long),
+                ('tv_nsec', ctypes.c_long)
+            ]
+
+        if sys.platform.startswith("linux"):
+            librt = ctypes.CDLL('librt.so.1', use_errno=True)
+            clock_gettime = librt.clock_gettime
+            clock_gettime.argtypes = [ctypes.c_int32, ctypes.POINTER(timespec)]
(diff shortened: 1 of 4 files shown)


mnaberez:
@Mook-as Thanks for your work on this.  I know the original motivation for this was #281, which I have hopefully fixed in eb96a2afa1cb9edf4db5c6bf3f1941ed9bb0b833.  It would be great if you were able to help test that the fix works.

I have been thinking about this PR, and I'm a bit nervous about the `ctypes` hacking that's needed to get this on Python 2.  As far as I can tell, Python 2 is a very large majority of our users and probably will be for some time, and they are all going to hit this code path.  It seems risky.  I'd be a lot more comfortable if we were able to just defer to Python stdlib for this.

The suggestion from @gjcarneiro is only use monotonic time if it is available.  How about something like this in `compat.py`:

    try:
        from time import monotonic as maybe_monotonic_time
    except ImportError:
        from time import time as maybe_monotonic_time

The `maybe_monotonic_time` name is ugly but it would make calling code clear that time might go backwards.  On Python 2, #281 would be fixed by eb96a2afa1cb9edf4db5c6bf3f1941ed9bb0b833.  On Python 3, `monotonic` would be used so eb96a2afa1cb9edf4db5c6bf3f1941ed9bb0b833 would be a no-op.  I'm not crazy about differing behavior across Python versions but it would add stability so it might be worth it.  What do you guys think?

mnaberez:
cc @msabramo @lukeweber You are guys are pretty active lately, what do you think about this PR?

mnaberez:
Closing for now since the motivation for this was #281 which should be fixed by eb96a2afa1cb9edf4db5c6bf3f1941ed9bb0b833.  We can reopen if discussion starts again.

Issue #109, opened 2012-04-25
Get rid of asynchat_25.py, asyncore_25.py and loggers.py


palmkevin:
Is there a special reason why supervisor/medusa is using asynchat_25.py, asyncore_25.py and loggers.py instead of their equivalent modules shipped with any standard python release? (asynchat, asyncore and logging)

Is it due to a lack of time of the supervisor developers that these are used or is there any other special reason?

My guess is that nobody wants to venture a too deep trip into medusa?!
BTW: Are there any plans to get rid of (or better: replace) medusa? This project seems quite dead?

evanunderscore:
Can we reopen this? Support for 2.5 is no longer a concern.

evanunderscore:
The incompatibilities mentioned make it non-trivial to swap out the bundled versions for the standard library versions.

evanunderscore:
I'm fairly convinced that medusa is at the root of a lot of the Python 3 porting issues. The package itself was written for Python 2, and we're trying to use it mostly as-is but with text_socket.py faking strings into and out of sockets, which breaks Python 3's `asynchat` and `asyncore`. Porting medusa to deal with bytes properly is going to be rough, since it has no tests (that I can see). Are there any newer alternatives that could be considered?

evanunderscore:
In the case that we decide to stick with medusa, here's my take on the steps needed if we were to properly port it to Python 3:

1. Write tests for medusa
2. Write tests for the pieces of supervisor that directly use medusa
3. Update bundled versions of asynchat/asyncore to include the Python 2.6 changes
4. Remove text_socket
5. Remove the bundled versions of asynchat/asyncore entirely
6. Update medusa to use bytes where sensible instead of native str
7. Update supervisor where necessary

1 and 2 are needed before any of this code is touched, and 3-7 all have somewhat interleaved dependencies on each other.

Given the amount of work this would take, it seems more sensible to me to replace medusa entirely. Either way this is more work than I'm willing to do at this stage, but hopefully this is of use to anyone else wanting to help out.

vsajip:
I think the above comments by @evanunderscore have nailed it - the use of `text_socket` seems like "the wrong fix", in particular. The `async{ore,hat}` modules seemed very neat when they first came to my notice, but they have some design flaws which I ran into when trying to extend `smtpd` for use in the stdlib tests for `logging`. These flaws limit the extensibility of the code therein.

[...]

PR #471, opened 2014-07-30
Fix tail -f not working on both supervisorctl and Web interface (ASCII and unicode log)

ngocson2vn:
I wrote a Python script for outputting `"%H:%M:%S"` each second. Then, I used tail -f function on both supervisorctl and Web interface. Unfortunately it didn't work

For supervisorctl, it output a data chunk including error message, then stopped:

```
supervisor> status
clock                            RUNNING   pid 21792, uptime 0:15:08
supervisor> tail -f clock
==> Press Ctrl-C to exit <==
b':29:33\n21:29:34\n21:29:35\n21:29:36\n21:29:37\n21:29:38\n21:29:39\n21:29:40\n21:29:41\n21:29:42\n21:29:43\n21:29:44\n21:29:45\n21:29:46\n21:29:47\n21:29:48\n21:29:49\n21:29:50\n21:29:51\n21:29:52\n21:29:53\n21:29:54\n21:29:55\n21:29:56\n21:29:57\n21:29:58\n21:29:59\n21:30:00\n21:30:01\n21:30:02\n21:30:03\n21:30:04\n21:30:05\n21:30:06\n21:30:07\n21:30:08\n21:30:09\n21:30:10\n21:30:11\n21:30:12\n21:30:13\n21:30:14\n21:30:15\n21:30:16\n21:30:17\n21:30:18\n21:30:19\n21:30:20\n21:30:21\n21:30:22\n21:30:23\n21:30:24\n21:30:25\n21:30:26\n21:30:27\n21:30:28\n21:30:29\n21:30:30\n21:30:31\n21:30:32\n21:30:33\n21:30:34\n21:30:35\n21:30:36\n21:30:37\n21:30:38\n21:30:39\n21:30:40\n21:30:41\n21:30:42\n21:30:43\n21:30:44\n21:30:45\n21:30:46\n21:30:47\n21:30:48\n21:30:49\n21:30:50\n21:30:51\n21:30:52\n21:30:53\n21:30:54\n21:30:55\n21:30:56\n21:30:57\n21:30:58\n21:30:59\n21:31:00\n21:31:01\n21:31:02\n21:31:03\n21:31:04\n21:31:05\n21:31:06\n21:31:07\n21:31:08\n21:31:09\n21:31:10\n21:31:11\n21:31:12\n21:31:13\n21:31:14\n21:3http://127.0.0.1:9001/logtail/clock/stdout Cannot connect, error: <class 'ValueError'> (invalid literal for int() with base 16: "1:15\\n21:31:16\\n21:31:17\\n21:31:18\\n21:31:19\\n21:31:20\\n21:31:21\\n21:31:22\\n21:31:23\\n21:31:24\\n21:31:25\\n21:31:26\\n'")
supervisor> 
```

Take a look at the output, you will see the error message:

```
http://127.0.0.1:9001/logtail/clock/stdout Cannot connect, error: <class 'ValueError'> (invalid literal for int() with base 16: "1:15\\n21:31:16\\n21:31:17\\n21:31:18\\n21:31:19\\n21:31:20\\n21:31:21\\n21:31:22\\n21:31:23\\n21:31:24\\n21:31:25\\n21:31:26\\n'"
```

For Web interface, sometimes it behaved same as supervisorctl, sometimes it didn't show any log.

With this fix, supervisor worked perfectly with Python-2.7.8 and Python-3.4.1


sontek:
+1 to this, it resolves: https://github.com/Supervisor/supervisor/issues/565

buzzijavier:
Whats going on with this?!

mnaberez:
@buzzijavier I actually don't know.  `supervisorctl tail -f` being completely broken on Python 3 will probably be unacceptable to users and would be a blocker to a release claiming Python 3 compatibility. I don't have much experience with Python 3 yet and I'd really like some people who have an understanding of Python 2 and 3 porting issues to review this code.  It may be that @ngocson2vn is correct, but I'm afraid I don't understand this code well enough yet to make that call, and I'd like some other eyes on it.  The concern I have is that the `supervisor.http` code being modified here is used for a lot more than `tail -f`, so changes to this module may destabilize other parts of Supervisor, even for Python 2 users.  I did try to draw some people to this issue and over the last few months I sent some private mail to people who were involved in the Python 3 changes to try and get some help but it hasn't happened.  For anyone else reading this, help would be appreciated.

msabramo:
@mnaberez: I'll try to take a look at this sometime this week.

supervisor/http_client.py
```diff
@@ -46,7 +50,7 @@ def __init__(
         ):
         asynchat.async_chat.__init__(self, conn, map)
         self.listener = listener
-        self.user_agent = 'Supervisor HTTP Client'
+        self.user_agent = SUPERVISOR_USER_AGENT
```
msabramo: This seems like a nice change but is it necessary in this PR? I'm wondering if maybe this should be a separate PR?

supervisor/http_client.py
```diff
@@ -132,12 +136,57 @@ def handle_connect(self):
...
-        self.listener.feed(self.url, data)
+        self.listener.feed(self.url, as_string(data))
 
-    def collect_incoming_data(self, bytes):
-        self.buffer = self.buffer + bytes
+    def collect_incoming_data(self, data):
+        if PY3:
+            self.buffer = as_string(self.buffer) + as_string(data)
+        else:
+            self.buffer += data
```
msabramo: I'm wondering if this should be:

~~~python
self.buffer = self.buffer + data
~~~

Because on my first look, I didn't notice the `+=` and was wondering for a bit why this wasn't including `self.buffer`, like it does in the Python 3 case. I like the idea of having it be a little more explicit so that the 2 branches look more similar.

supervisor/http_client.py
```diff
@@ -201,7 +250,10 @@ def chunked_size(self):
         line = self.buffer
         if not line:
             return
-        chunk_size = int(line.split()[0], 16)
+        try:
+            chunk_size = int(line.split()[0], 16)
```
msabramo: Silently swallowing an exception always makes me nervous. What's the purpose here? Maybe add a comment?

supervisor/http_client.py
ngocson2vn: Yes, I agree

supervisor/http_client.py
ngocson2vn: Actually we don't need try ... except here. I'll remove it.

supervisor/http_client.py
ngocson2vn: Yes, I agree.

waldyrious:
FWIW, now that [python-rsa](http://stuvel.eu/rsa)'s red state has been worked out through [issue #7](https://bitbucket.org/sybren/python-rsa/issues/7/fix-python3-support), supervisor is the [top red entry in the python 3 wos](http://python3wos.appspot.com/). What's holding this PR up, and what other issues need resolving before releasing a version with Python3 support? I wish I could help, but I'm just an unexperienced enthusiast :(

ngocson2vn:
Is there anymore feedback on this patch @mnaberez @msabramo?

msabramo:
Has conflicts now, but otherwise this is looking good to me.

supervisor/http.py
```diff
@@ -663,7 +659,7 @@ def more(self):
             self.file.seek(-bytes_added, 2)
             bytes = self.file.read(bytes_added)
             self.sz = newsz
-            return bytes
+            return as_string(bytes)
```
regebro: Although I don't mind the other changes, they may or may not be necessary, it is in fact this change that solves the actual bug.

regebro:
Does anyone have a good idea for how to write unittests for this error?

supervisor/http.py
ngocson2vn: @regebro: the other changes are aimed to display unicode log. Please help to review them too!

regebro: Yup, I can reproduce that error as well, and although ugly and converting between strings and bytes at multiple places, the option (using bytes everywhere) requires massive changes everywhere including in medusa. I think this may be teh best path forward. Could you update this pull request to reflect the other changes discussed with msabramo?

lukeweber:
Fixed the merge conflict, and removed the try catch that was mentioned above. I've verified that tail and tail -f work inside supervisorctl at command prompt and put it together here: https://github.com/lukeweber/supervisor/tree/py3-fixes

Also encoding headers, and meta tags aren't correct in web and things encoding was therefore wrong in the web views:
https://github.com/lukeweber/supervisor/commit/1c62548faf979d0654898ed137dd7cbe93c0b532

My test program:
```sh
#!/bin/bash

while :;
    do echo "halloü";
    echo "hallo";
    sleep 1;
done
```

I did notice that "tail -f" in the web doesn't seem to work(no response) until the output wraps(wait a while) and looks like:
```
o
halloü
hallo
halloü
hallo
halloü
hallo
```

I don't know enough about this code yet to know what the difference is. The must be an issue about not being able output less than a more than full log with the new code in the tail -f web code.

bgruening:
Any news about this one?

razvanphp:
I just want to mention that this is an issue even in Python 2.7.9, over http web interface.

Merging / fixing this 2 years old bug would be awesome!

ngocson2vn:
@razvanphp I will resolve conflicts soon.

bgruening:
ping @mnaberez @mcdonc
Thanks  @ngocson2vn for the rebase!

zfrenchee:
@mnaberez @mcdonc Any news on this? This is important to me.

mnaberez:
My reading of the tickets and this PR is that there are several separate issues around the tail function on the web interface.  My understanding of this PR is that it attempts to raise the unreleased Python 3 port to the same level of broken as the released Python 2 version.  In other words, merging this might improve unreleased Python 3 support, but probably isn't going to help issues people have having with the released versions of Supervisor on Python 2.

bgruening:
It's an improvement as you said :) Anything that needs to be done to get this merged - or is this going to be closed? Either way a decision should be taken here so that we all can go on.

mnaberez:
> Anything that needs to be done to get this merged - or is this going to be closed?

Supervisor has some [major issues](https://github.com/Supervisor/supervisor/issues?q=is%3Aissue+is%3Aopen+%22python+3%22+label%3A%22python+3%22) on Python 3.  Many of these issues are encoding related as well so merging this one patch doesn't move the needle much.  We need someone who has strong experience in Python 2/3 porting and is willing to spend a non-trivial amount of time looking at these bytes/strings issues together.  This patch does some things that suggest there might be a better solution, e.g. it overrides a large portion of asynchat and converting from bytes to string only to convert back again.  Hopefully there is a cleaner solution, but perhaps not.

> Either way a decision should be taken here so that we all can go on.

As far as closing it for the sake of closing it, I don't think we should do that.  The Python 3 issues should actually be fixed rather than just closing them.

sandys:
@mnaberez can you make this a formal call for volunteers and put it up as a blog post or something over at hacker news ?

I dont think people even know that supervisord project needs help for porting from python 2 to 3

mnaberez:
This was done already on the mailing list and has been in the README for a couple of years.

nicktimko:
Been looking through some of this as per [call at Reddit](https://www.reddit.com/r/Python/comments/5k1875/supervisord_project_python_2_to_3_porting_help/). I've ported and written various 2/3 compatible libraries, and the initial reaction is that holding on to Python 3.2 support makes it extremely annoying because it doesn't support `u''` syntax and it was [readded in 3.3](https://docs.python.org/dev/whatsnew/3.3.html). 2.7 support has entirely bracketed 3.2:

* [Python 2.7.0 final: July 2010](https://www.python.org/dev/peps/pep-0373/)
* [Python 3.2.0 final: February 2011](https://www.python.org/dev/peps/pep-0392/)
* Python 3.2 EOL: February 2016
* Python 2.7 EOL: >=2020? 

I'd hope any platforms that only supported one of 2 and 3 would at least hold on to 2.7. As there is no current support for 3.x, can the notion of 3.2 support just be dropped?

evanunderscore:
I second the idea of dropping support for 3.2. Most projects now seem to be going for 2.7 and 3.3+, so unless you have a compelling reason this needs to run on 3.2, it's more hassle than it's worth.

My 2 cents on the porting effort - `as_bytes` and `as_str` are doing more harm than good. In general, you should know whether your data is textual or binary most of the time, and only use these kinds of conditional type conversions when you really can't be sure. There should almost definitely never be an `as_bytes(self.buffer)` because the object should know what type its buffer is (it should already be bytes).

The `unicode` definition [here](https://github.com/Supervisor/supervisor/blob/27ac565b282fe5c5772d589f7c6255622a5ee8c8/supervisor/compat.py#L11) must be causing bugs in the handful of places it's used ([here](https://github.com/Supervisor/supervisor/blob/219f7af0ac8b0b1965b00793a923a22addfadbaa/supervisor/rpcinterface.py#L906) and [here](https://github.com/Supervisor/supervisor/blob/427eb2bc6b08f788573deb91d1391d93f8b58a1b/supervisor/supervisorctl.py#L263)) because the type checks will fail. I tried replacing this with `unicode = str` and it broke a few tests, most notably [this](https://github.com/Supervisor/supervisor/blob/27ac565b282fe5c5772d589f7c6255622a5ee8c8/supervisor/tests/test_rpcinterfaces.py#L2042) which I think is bad - these should either be textual or data in both versions. Given that the input is textual, I'm guessing the output should be too. Someone more familiar with the code might be able to comment.

Apologies for just dumping everything on this MR - is there a more appropriate place for general Python 3 discussion?

nicktimko:
It seems like all the tests pass on master (dropping Python 3.2), is the problem in something that's not covered or some compound code path (do this, then this, then this...).

* https://travis-ci.org/nicktimko/supervisor
* https://codecov.io/gh/nicktimko/supervisor/tree/2to3no32/supervisor

evanunderscore:
Are you asking about the `unicode` definition problem? In Python 3, nothing is ever an instance of that class so those tests are always returning False. The bits of code I linked can't be doing what they intended. They only appear to be hit in the coverage report because they work on Python 2. Whether this manifests somewhere as an actual bug or everything cancels out because `as_bytes` and `as_str` are used everywhere, I don't know. The tests pass because they assert different outputs for Python 2 and Python 3.

evanunderscore:
Potentially also relevant: #109

ztane:
Python 3.2 was totally EOLed a year ago, it gets no fixes, same for 2.6. Those who are stuck with 2.6 or 3.2 can be stuck with older supervisor as well, as they don't seem to care about new features and bugfixes.

pgerhardt:
Having this issue with 3.3.4 on #1029

radome:
Having this issue with 4.1.0 Python 3.5.6 on any browser.

DespLegion:
Having this issue with 4.2.1 Python 3.10 on any browser

xzxiaoshan:
Having this issue with 4.2.5 Python 3.13 on Chrome browser

Issue #179, opened 2012-12-04
Send a generic signal to a process


issackelly:
I would suggest `supervisorctl signal <process group> <signal>`  so, `supervisorctl signal gunicorn HUP`  

Relates to #166 and #53

Issue #53, opened 2011-11-22
Restart signals not supported


ido50:
Hi there. Thanks a lot for supervisor, I am using it on quite a few servers and it is performing greatly.

I would like to suggest a feature that would be very advantageous (well, at least for me). There are some processes that support receiving a certain signal, such as HUP, causing them to perform a graceful restart.

Examples for such processes include the HTTP web server Cherokee (http://www.cherokee-project.com) and servers initiated via Perl's Server::Starter (http://search.cpan.org/dist/Server-Starter), which acts as a superdaemon.

If I could set on supervisord.conf that a certain process is restarted with a HUP signal, just like I can set the signal for stopping a process, then I could just do "restart process_name" and get the correct behaviour. That would be awesome.

Thanks a lot,
Ido Perlmuter.

lonetwin:
+1 on this request, irrespective of implementation details. I am interested in using the USR2 signal facility for nginx ( http://nginx.org/en/docs/control.html#upgrade ) (yes, we run nginx under supervisord !) to do on the fly nginx upgrades. Unfortunately this currently doesn't work since sending a USR2 signal to nginx (ie: outside of supervisord context) creates a new nginx master process which is not a child of supervisord.

Feng23:
+1 Hope to reload gracefully

Feng23:
Finally I set the ```stopwaitsecs``` long enough to work around my process. 
The number of seconds to wait for the OS to return a SIGCHILD to supervisord after the program has been sent a stopsignal. If this number of seconds elapses before supervisord receives a SIGCHILD from the process, supervisord will attempt to kill it with a final SIGKILL.

leandrosouza:
apt-get install supervisor
My version is:
# dpkg -l |grep -i supervisor
ii  supervisor                          3.0b2-1

 - restart the program 
# supervisorctl restart program-name
i.e.

# supervisorctl restart order-form
order-form: stopped
order-form: started

zllak:
+1 for graceful reload, where we can actually choose the graceful signal to send :)

andy-pham:
+1

munzur:
+1

cades:
+1

hakanw:
+1

bijenkins:
+1

CVirus:
+1

staab:
There are 390 forks for this. Does anyone know if there is one that implements this change?

surfeurX:
+1

[...]

PR #477, opened 2014-08-14
Add ability to send arbitrary UNIX signals to processes:

msabramo:
- Extend the Process class to handle non-killing signals
- Expose this via the XML-RPC API
- Add supervisorctl `signal` command

This is an update of PR #228 and hopefully moves the ball forward on #179, #53, etc.

Cc: @squeed, @weebly, @mnaberez, @sontek

### DIFF
--- a/supervisor/rpcinterface.py
+++ b/supervisor/rpcinterface.py
@@ -497,6 +497,8 @@ class SupervisorNamespaceRPCInterface:
         if not msg is None:
             raise RPCError(Faults.FAILED, msg)
 
+        return True
+
 
     def sendGroupSignal(self, name, signal='HUP'):
         """ Send a signal to all processes in the group named 'name'
--- a/supervisor/tests/test_rpcinterfaces.py
+++ b/supervisor/tests/test_rpcinterfaces.py
@@ -847,7 +847,7 @@ class SupervisorNamespaceXMLRPCInterfaceTests(TestBase):
         supervisord.set_procattr('foo', 'state', ProcessStates.RUNNING)
         interface = self._makeOne(supervisord)
 
-        result = interface.sendProcessSignal('foo', 10)()
+        result = interface.sendProcessSignal('foo', 10)
 
         self.assertEqual(interface.update_text, 'sendProcessSignal')
         self.assertEqual(result, True)


IvanAnishchuk:
> it simply does nothing and issues a warning
> So I would add a "--force" option to the graceful command which falls back to stop/start if no graceful signal is defined.

I believe that should be the default behavior with option to do nothing if there's no graceful signal defined (`--calm`?). And warning could be issued while doing the job, I don't think anybody would go and edit the config when they need to restart something and `--force` option will only increase the command line required for the most needed function (and re-sending rate, sometimes causing unneeded consecutive restart).

lonetwin:
Just curious what version of supervisor are this changes set to be released in ?

mnaberez:
The `signal` command from this pull request will be included in the next release, 3.2.

jgadling:
Any ETA on the 3.2 release?

astlock:
Thank you for the signal implementation! I found some wrong behaviour with signals and with normal restarts too:

```
[koshevoy@koshevoy-staging.dal05 ~]$ supervisorctl signal sigusr2 app-api
app-api: ERROR (no such process)
[koshevoy@koshevoy-staging.dal05 ~]$ echo $?
0
```

```
[koshevoy@koshevoy-staging.dal05 ~]$ supervisorctl restart elf
elf: ERROR (no such process)
elf: ERROR (no such process)
[koshevoy@koshevoy-staging.dal05 ~]$ echo $?
0
```
If the restart or signaling failed for any reason, there should be exit code 1. Am I right?

mnaberez:
> If the restart or signaling failed for any reason, there should be exit code 1. Am I right?

The exit status always being 0 is not limited to signals.  It is tracked as #24 and there is now a pull request to change this in #688 that should be merged for a future release.

astlock:
I missed this issue. Thank you for the explanation!

mangelozzi:
This sounds like a great way to use `nginx -s reload` to reload the nginx config with zero downtime. Unfortuantely the [only help I could find on it says](http://supervisord.org/running.html#running-supervisorctl):

> signal
>     No help on signal

Say I have a program called nginx, how would I send the signal to nginx to reload its config?

Issue #29, opened 2011-07-15
Web Interface Behind Proxies


mnaberez:
There should be some way to configure a URL prefix for when the web interface is behind a proxy.  This has been requested several times on the mailing list.

Related to issue #28

mnaberez:
Reopening since this has come up again.  Related: #495

calvin:
I'm updating the code to support this. I know that there are other implementations (#495 and #28) but what I'm proposing is similar to #372. Once I'm done, I'll create a PR.

yarikdot:
btw: This nginx configuration worsk for me fine:

    location /supervisor {
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        # hack the host https://github.com/Supervisor/supervisor/issues/251
        proxy_set_header Host $http_host/supervisor/index.html;
        proxy_redirect off;
        rewrite ^/supervisor(.*)$ /$1 break;
        proxy_pass http://127.0.0.1:8999/;
    }

archon810:
@yarikdot This isn't working for me, at least on v3.3.3, because it's trying to load http://MY_HOST/stylesheets/supervisor.css which is a 404. But even http://MY_HOST/supervisor/stylesheets/supervisor.css is a 404 too.

bfx:
@yarikdot works for me on v3.2, but not the "tail -f" view of the logs

Any progress on pull request: https://github.com/Supervisor/supervisor/pull/593

potoo0:
`base_path` option have been discard ?

mwansinck:
Love to see a URL prefix being added, any update on this?
PR #593 seems OK to me.

rosenhagen:
One detail that seems to add to the problem is the missing path in the "Location" header. Maybe there would be no URL prefix necessary in most cases if one was present?

For example, this is a location header that I would typically get:

    Location: http://supervisor:9001?message=…

Please note: no path or "/" present. It's a perfectly valid URL, but in my case the missing slash messes with my proxying in Apache, which should be able to modify the URL out of the box.

To get this working I had to use a rather unusal config:

```
ProxyPass /supervisor/ http://127.0.0.1:9001/
ProxyPassReverse /supervisor/ http://127.0.0.1:9001
```

… e.g. trailing sash on request, no slash on responses. I'm not sure if the missing path is a bug, it seems at least unusual and think one should be added.

PR #495, opened 2014-09-24
Fix redirect with urlrewrite setup

Guria:
Redirect must be relative to support running with url rewrite rules

### DIFF
--- a/supervisor/web.py
+++ b/supervisor/web.py
@@ -429,8 +429,7 @@ class StatusView(MeldView):
                 if message is NOT_DONE_YET:
                     return NOT_DONE_YET
                 if message is not None:
-                    server_url = form['SERVER_URL']
-                    location = server_url + '?message=%s' % urllib.quote(
+                    location = '?message=%s' % urllib.quote(
                         message)
                     response['headers']['Location'] = location
 


mnaberez:
Thanks for the pull request but I have decided not to merge it at this time.  As far as I can tell, we've never advertised that the web interface will work in a URL rewrite setup.  I'd rather not make this change because once we say we support it, we can't easily take that away.  The web interface could use some improvement, but supporting it on any arbitrary URL will make that development more difficult.

Guria:
It is not about url rewrite support. Not all environments provide ability to run any tools on any port you want (NAT). So there is only 2 options to use web-uis for tools: subdomains (I'm afraid not supported in such environments as well) and placing them on "subfolder" of main site. You should not expect that web-ui always placed under the root of site.

mnaberez:
> It is not about url rewrite support.

Sorry I was confused by both the title and description of the patch then.

Guria:
Url rewrite is just technic used to place one web-app in sub-path of domain. Sorry, I had to provide more detailed explanation in PR.

mnaberez:
There's now two proposed solutions (#28, this one).  I've reopened #29.

PR #500, opened 2014-10-14
Change spawned log info as group name and process name. Issue 452

gjcarneiro:
I fixed kozdincer's earlier patch, I hope (we'll see from travisCI).


razvanphp:
@gjcarneiro can you please resolve the merge conflicts? I would really appreciate this fix, 2 years it's a long time for such a small but useful change. Thank you!

Issue #487, opened 2014-09-05
feature request: delay restart before status FATAL


guettli:
Use case:

during a software update the server can't restart since some needed python modules are not available yet. Supervisor retries N times (N comes from config. AFAIK defaults to 3).

After N fails the state FATAL gets entered.

In my use case the programm entered state FATAL although just some seconds later the restart would have been successfull.

Can you understand this use case?

A possible solution would be to use a delay in unit "seconds": After N failed retries wait M seconds and then try again. In my case this should be done to infinity (endless loop).

aztlan2k:
:+1:   this would be really useful.

allowing us to configure how long between retries would be great.  Allowing for an backoff would also help.    

i think it would be great if we could also specify a retry count of infinity (never give up) in combination with a sane value for the `delay_betwen_retries`.

mmh:
+1, would be useful.

willybarro:
:+1: would be great. There's already a PR with that feature https://github.com/Supervisor/supervisor/pull/509, but it's not automergeable, and there's no discussion on it.

mrook:
+1

nicwest:
+1

arinto:
+1

nereusz:
+1 This would be very useful for me

x0e-foobar:
+1

MatthiasBach:
+1

jsmirnov:
+1

guettli:
Has someone enough knowledge to create patch for this?

dbpolito:
I would love to see this too, so :+1:

detailyang:
+1

oryband:
:+1:

vvanscherpenseel:
Would love to have this feature. :+1:

fordnox:
+1

siavashs:
👍

guettli:
Why was this ticket closed? Was the feature request implemented? I can't see a code change on this issue or on #561.

tonicospinelli:
+1

oryband:
dear sir plz implement this alrdy k thx bai

eggsby:
:thumbsup: would like this feature, thanks!

0x20h:
+1, would be useful for me too!

caorong:
+1, hope this!

conanfanli:
+1

dieend:
+1

miso-belica:
[...]

PR #509, opened 2014-11-01
Added a "restartpause" option

fclairamb:
This allows to set a pause between the restarts of the program.

Typically, when you restart a database server, this can take few minutes and the backoff delay won't be enough. With a restart pause delay of something like 30 seconds, this becomes a lot less likely. 
The alternative is to increase the number of start retries, but on system startup and/or when many programs fail at the same time (usually for the same reason), it creates an unnecessary workload on the host.

This changes the behaviors in two scenarios:
* Applied on startup failure additionally to the backoff delay
* Applied on bad exit status

Default value is 0, which doesn't change the existing behavior.

It is mostly a proposal at this stage. We could also allow to change the backoff increment (which is hardcoded to 1 at the current stage) or only the backoff base delay.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -848,6 +848,7 @@ class ServerOptions(Options):
         autorestart = auto_restart(get(section, 'autorestart', 'unexpected'))
         startsecs = integer(get(section, 'startsecs', 1))
         startretries = integer(get(section, 'startretries', 3))
+        restartpause = integer(get(section, 'restartpause', 0))
         stopsignal = signal_number(get(section, 'stopsignal', 'TERM'))
         stopwaitsecs = integer(get(section, 'stopwaitsecs', 10))
         stopasgroup = boolean(get(section, 'stopasgroup', 'false'))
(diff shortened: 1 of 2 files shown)


oryband:
:+1:

willybarro:
@mnaberez @mcdonc Hey guys, any reason (asides from time) for this pull request not being accepted/reviewed? There's another thread with over 15 :+1: guys wanting this feature badly.

oryband:
@ofir-petrushka @fuzzyami @doodyparizada

doodyparizada:
:+1:

mrook:
+1

miso-belica:
@Supervisor is there any problem with this PR. There is also #487 where people ask for this. Is there something I can help with?

pfuender:
PR #659 is based on your work, fclairamb and includes fixed for the tests. Wish me luck ;-)

fclairamb:
Thank you very much !!!

briandignan:
Merge please? :'(

PR #512, opened 2014-11-12
Add grace time option to pidproxy

aebm:
The grace time option allows to have a maximum wait time for process
termination. If the process doesn't end in that time frame pidproxy
will send the kill signal.

Credits to @gjcarneiro for the comments in the sleep section

### DIFF
--- a/supervisor/pidproxy.py
+++ b/supervisor/pidproxy.py
@@ -13,7 +13,14 @@ class PidProxy:
     def __init__(self, args):
         self.setsignals()
         try:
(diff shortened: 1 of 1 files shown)


mnaberez:
Thanks for this pull request, but I don't think we are going to add more options to `pidproxy`.  We are including `pidproxy` in the Superivsor package mainly for historic and compatibility reasons.  It would probably be better if new/improved `pidproxy` programs were their own package and we listed them on the [plugins](https://github.com/Supervisor/supervisor/blob/3.3.3/docs/plugins.rst) page.

PR #528, opened 2014-12-02
supervisorctl read from ~/.supervisorctl.conf

msabramo:
This is a good place for users to do user-specific customizations, like change the prompt or add plug-ins.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -110,6 +110,11 @@ class Options:
                        'supervisord.conf',
                        'etc/supervisord.conf',
                        '/etc/supervisord.conf']
+        if self.progname.endswith('supervisorctl'):
+            dot_supervisorctl_conf = os.path.join(
+                os.path.expanduser('~'),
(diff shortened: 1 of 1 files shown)


hobarrera:
I'd rather follow the xdg-basedir-spec and read per-user settings from `$XDG_CONFIG_DIR/supervisord/supervisorctl.conf` (`~/.config/supervisord/supervisorctl.conf` by default) I actually had a similar patch in mind for supervisord in mind as well already.

mnaberez:
I've left some comments about the search paths over in #576.  I would welcome discussion about this and other search path issues in that thread.

mnaberez:
I'm going to close this in favor of #742.  Several people now have requested letting the config filename be set by an environment variable so I've come around on it.  If that was implemented, this PR would not be needed, since you could put `export SUPERVISOR_CONF=$HOME/supervisord.conf` in your `~/.profile`.

PR #530, opened 2014-12-03
Colorized output for supervisorctl status

msabramo:
How about a sexy new feature for 4.0...? :smile: 

I've long wanted this, because it makes the failed processes stand out from the OK ones in a long list.

supervisorctl_status_colored

I'm toying with the idea of colorizing the entire line rather than just the status. Or maybe the name and the status. Having the name colorized seems nice so that your eyes can just scan the first column. Kind of thinking of not colorizing the rest of the line according to status, because I had an idea of maybe colorizing the uptime if it less than a certain threshold (`startsecs`?) -- this is because we sometimes have processes that repeatedly spin up and die and use a lot of CPU, so it can be very useful sometimes to spot the processes that have the shortest uptime...

@mcdonc and @mnaberez: What do you guys think?

I didn't write tests, but I will add them if you're cool with the feature.

### DIFF
--- a/supervisor/colors.py
+++ b/supervisor/colors.py
@@ -1,0 +1,31 @@
+ESC = '\033['
+
+
+class ansicolors:
+    class fg:
+        black = ESC + '30m'
+        red = ESC + '31m'
+        green = ESC + '32m'
+        yellow = ESC + '33m'
+        blue = ESC + '34m'
+        magenta = ESC + '35m'
+        cyan = ESC + '36m'
+        white = ESC + '37m'
+        reset = ESC + '39m'
+
+    class bg:
+        black = ESC + '40m'
+        red = ESC + '41m'
+        green = ESC + '42m'
+        yellow = ESC + '43m'
+        blue = ESC + '44m'
+        magenta = ESC + '45m'
+        cyan = ESC + '46m'
+        white = ESC + '47m'
+        reset = ESC + '49m'
+
+    class style:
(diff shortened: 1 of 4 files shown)


gjcarneiro:
+1

msabramo:
I'll try to come back to this in a few days. Any changes/ideas?

jvrplmlmn:
The colorized output looks like a great addition :+1:

msabramo:
Squashed it down to a single commit.

eggsby:
+1

mnaberez:
Thanks for updating this patch to remove the dependency.  I think this is a nice addition and I would be happy to merge this if some test coverage was added.  

I looked through the new patch and there's one thing I would like to discuss: Right now the option to use color is specific to the status command.  Should we add one "use color" option that affects all of supervisorctl, or is there a use case for specifying color/no color per command?  

There might be other uses for color, e.g. we might want to show errors in red, or plugins might want to colorize something.  I'm wondering if we will accumulate color options for those cases or if one global "use color" option would be sufficient.

msabramo:
Interesting. I had only thought of using color for `status` and hadn't thought of other uses, but that's a good point -- perhaps there should be one global color option.

dirkaholic:
Why was this PR never considered to be merged ? I find the feature mega useful.

FAKERINHEART:
when the target to output is not a terminal, such as pipe, I think it is better not to output colorfully.

PR #552, opened 2015-01-08
fix _assertInState error when system time changed and one child process exit ...

samhair:
...itself


mnaberez:
Related: #533

mnaberez:
This issue is identical to #533 and the proposed fix is almost the same as in that one also.  I'm going to close this as a duplicate, please leave a comment in #533 if you see any issues with that patch vs. this one.

PR #567, opened 2015-02-01
Attempt to fix %(here)s in included files. Fix #81, related to #84

### LINKED ISSUE #81: Make a counterpart of %(here)s for included configs

alexef:
I've been giving some thought and investigation and because of how ConfigParser is implemented (in 2.7 and 3.4) I couldn't come with more elegant solution.

Suggestions are welcome.

I'm using string replace instead of format because the same value can have multiple substitutions, for example: 
xx = %(here)s/%(program_name)s

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -567,6 +567,7 @@ class ServerOptions(Options):
         expansions = {'here':self.here}
         expansions.update(self.environ_expansions)
         if parser.has_section('include'):
+            parser.expand_here(self.here)
             if not parser.has_option('include', 'files'):
                 raise ValueError(".ini file has [include] section, but no "
                 "files setting")
@@ -591,6 +592,10 @@ class ServerOptions(Options):
                         parser.read(filename)
                     except ConfigParser.ParsingError as why:
                         raise ValueError(str(why))
+                    else:
+                        parser.expand_here(
+                            os.path.abspath(os.path.dirname(filename))
+                        )
 
         sections = parser.sections()
         if not 'supervisord' in sections:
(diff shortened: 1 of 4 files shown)


mnaberez:
Thanks!  I agree and hopefully at some point we can come up with a more elegant way of fixing this, but in the meantime, the behavior of `%(here)s` should be fixed.  I've merged this patch and we can always update it later if we figure out a better way.

geekdevs:
I don't understand how releases work for this project. This commit hit master almost a year ago, latest release was ~1 month ago yet this change is not in that release? Really?

mnaberez:
This is a breaking change so it will be included in the next major version.

PR #568, opened 2015-02-01
program start delay option (on first start only)

liutec:
Only applies startdelay when the supervised program is started for the first time. The default value of 0 seconds fully preserves BC.


supervisor/process.py
```diff
@@ -209,6 +211,9 @@ def spawn(self):
         self._assertInState(ProcessStates.EXITED, ProcessStates.FATAL,
                             ProcessStates.BACKOFF, ProcessStates.STOPPED)
 
+        if delaystart:
+            time.sleep(self.config.startdelay)
```
mnaberez: Supervisor is an asynchronous program.  Blocking like this is unacceptable.

That aside, I'm not sure any kind of delay-before-something features need to be added to Supervisor.  You can have Supervisor run a shell script with some variation of this easily enough:

    #!/bin/sh
    sleep 5
    exec your-program-here

liutec: I could not agree more. Unfortunately the extent of modifications required to implement this in an asynchronous manner will never get accepted. With this ugly hack I've just made a poorly implemented feature which desperate people like myself may use to quickly prevent machine crashes when starting all supervised programs (because they all get started at the exact same time).

As for your example: that will only delay my problem 5 seconds.
Because I need each program to get started N seconds after the previous one was started. Eg: 3 programs with a 5 second start delay: prog. 1 at 12:00:00, prog. 2 at 12:00:05 and prog 3 at 12:00:10 - this will distribute the startup peak load.

liutec: The startdelay gets compounded for every process, (proc n gets a n*startdelay seconds delay) and it only happens the first time it starts (I don't believe it will mess with the async operation after startup)

mnaberez: > I need each program to get started N seconds after the previous one was started.

Another shell script solution would be something like this:

    #!/bin/sh
    supervisorctl start program1
    sleep 5
    supervisorctl start program2
    sleep 5
    supervisorctl start program3

You would then use this script to start them rather than `supervisorctl start all`.

It is also possible to add your own commands to `supervisorctl`.  You could add a custom start command.  It's fine to block inside `supervisorctl`, you just can't block inside `supervisord`.  I have an [example project](https://github.com/mnaberez/supervisor_cache) that shows how to add a command to `supervisorctl`, it's called a `ctlplugin`.

mnaberez: > I don't believe it will mess with the async operation after startup

It is not acceptable to block inside `supervisord` under any circumstance.

liutec: Thank you, I will look into that. My first attempt was to only block in supervisorctl but there, as I remember, I did not have access to the individual processes for programs having num_procs > 1.

mnaberez: You can control the individual processes with `supervisorctl`.

With this program config:

    [program:foo]
    command=/bin/foo
    process_name=foo_%(process_num)s
    numprocs=2

Start all processes in the `foo` group:

    $ supervisorctl start foo:

Start individual `foo` processes:

    $ supervisorctl start foo:foo_0
    $ supervisorctl start foo:foo_1

mnaberez:
Closing for now.

PR #596, opened 2015-03-23
bdist_rpm: Use correct RPM name for meld3

KristianLyng:
The package on CentOS/Fedora/RHEL is "python-meld3", not just "meld3". This
commit fixes the name so the resultant RPM will be installable using
standard packages.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,7 +9,7 @@ cover-erase=1
 
 [bdist_rpm]
 requires = 
-  meld3
+  python-meld3
 
 [aliases]
 dev = develop easy_install supervisor[testing]


mnaberez:
@davidbirdsong Can you please review this change?  This changes the original `bdist_rpm` section that you added in #230.  

I am not an RPM user but I am wondering if the original version (`meld3` not `python-meld3`) was intentional.  The `bdist_rpm` section exists only for people who want to build their own RPM packages of Supervisor.  The reason people would do that would be to get the latest version of Supervisor instead of an older distribution-packaged version.  Those same people probably want to make their own RPM package from our latest [meld3](https://github.com/supervisor/meld3) version also.

davidbirdsong:
@mnaberez It's been a long time and I'm no longer on any RHEL distros to be sure. I think I recall having to build and maintain my own meld3 to use the later one and to build and maintain my own, up-to-date supervisor on my own repo.

Issue #286, opened 2013-08-31
Supervisor should allow setting a tag for syslog logging


mohit:
Logs in syslog can have a tag, that is often used to identify process names. Supervisor logs contain the process name but this is not recognized as the process name, and is always supervisor. 

While this works for simple cases, it becomes tiresome to filter different processes when managing many of them using supervisor. 

For an example of how this should look, please see (in a debian/ubuntu system):

```
echo "test syslog message" | logger -t "test_app"
```

and then check /var/log/syslog

honzik666:
This feature would be highly appreciated, I really miss it.

sukrit007:
+1 for this fix. If no one is working on this , I can take a stab at providing a fix.

Vye:
I'm not working on it but could still use this functionality.

--Vye

jlbribeiro:
:+1: for this feature.

till:
Apologies for maybe commenting on the wrong issue (there are too many threads (issues/PRs) about syslog and supervisord on here), we are after this feature as well.

In addition, I would like to make the log lines compatible to the standard syslog format.

I would be happy to send PRs to this, but I am not sure what to base my work on (implying that I would hope it's integrated or possible to integrate it relatively soon).

Any pointers?

PR #289, opened 2013-09-16
Better syslog configuration and output

samv:
Hi there, thanks for the recent rework in the syslog support.  This is a set of changes which make the syslog output a bit better, reaching this:

```
    Sep 19 10:18:46.285916 Herman daemon.notice supervisord[5296]: supervisord started with pid 5296
    Sep 19 10:18:47.288994 Herman daemon.notice supervisord[5296]: spawned: 'vmstat' with pid 5299
    Sep 19 10:18:47.291366 Herman daemon.notice supervisord[5296]: spawned: 'badvmstat' with pid 5300
    Sep 19 10:18:47.297447 Herman   user.info   vmstat: procs -----------memory---------- ---swap-- -----io---- -system-- ----cpu----
    Sep 19 10:18:47.298812 Herman   user.info   vmstat:  r  b   swpd   free   buff  cache   si   so    bi    bo   in   cs us sy id wa
    Sep 19 10:18:47.299112 Herman   user.info   vmstat:  3  0      0 4874284 244368 1695416    0    0    63    20   62  191  1  0 98  1
    Sep 19 10:18:47.299428 Herman   user.notice badvmstat: vmstat: failed to parse argument: '5;'
    Sep 19 10:18:47.299866 Herman daemon.notice supervisord[5296]: exited: badvmstat (exit status 1; not expected)
    Sep 19 10:18:48.298243 Herman   user.info   vmstat:  0  0      0 4863852 244368 1702904    0    0     0     0 1437 8074  4  3 93  0
    Sep 19 10:18:48.298447 Herman daemon.notice supervisord[5296]: success: vmstat entered RUNNING state, process has stayed up for > than 1 seconds (startse
    Sep 19 10:18:49.298594 Herman   user.info   vmstat:  0  0      0 4863528 244368 1703152    0    0     0     0  737 1475  1  1 99  0
      ...
    Sep 19 10:19:13.488894 Herman daemon.warning supervisord[5296]: received SIGINT indicating exit request
```

Along the lines of the new "stdout_syslog" and "stderr_syslog" boolean values, I have created a new "syslog" top-level configuration, which is configured identically.  Additionally, there are now reasonable defaults for syslog facility and priority (shown in the above, but not logged by default by eg rsyslog).  Finally, if you care, you can select the syslog channel (192 to choose from!) yourself for each stream by putting the channel there instead of "true" in any of the `*syslog` config settings.

Closes issue #162

PR #604, opened 2015-04-06
Support for syslog tag

sukrit007:
Introduce stdout_identifer , stderr_identifier that can be used as identifier (tag) for syslog

Possibly fixes: https://github.com/Supervisor/supervisor/issues/286

example config 
```
[program:test-syslog-tag]
command=/bin/bash -c "while true; do echo test message($(date))... ; sleep 5s; done "
stderr_logfile=syslog
stdout_logfile=syslog
stdout_identifier=my-tag
stderr_identifier=my-tag
```

If the tag is not specified , it defaults to 'supervisor'  in order to maintain existing behavior.


mnaberez:
I'm going to close this because it's against the 3.1-branch.  Please open feature requests against the master branch.  There's another pull request in #289 for syslog improvements.  It would be great if you could try to  work with @samv and @jaraco so we don't have competing syslog patches.

PR #611, opened 2015-04-28
Minor code cleanup.

aleksa:
- Fixed an `if()` comparison that should have used `None`.
  - Changed `while 1` to `while True`.
  - Replaced throwaway list comprehension with `for` loop, to avoid generating extra list in memory.


mnaberez:
No thanks. These changes are pretty arbitrary.

Issue #322, opened 2013-11-25
Feature request: Have an option to compress log files after their rotation


nilstoedtmann:
I know that there are tools like logrotate that can do rotate-and-compress [1]. But logrotate cannot enforce {streamname} _logfile _maxbytes just-in-time [2]. Therefore i preferred supervisor to handle the rotation.

---

[1] https://lists.supervisord.org/pipermail/supervisor-users/2010-November/000751.html 
[2] I have processes that occasionally, e.g. if certain errors occur, fill logs rapidly. By the time logrotate runs next time, the limited storage on my embedded system is already full.

danslimmon:
+1

miliauskaite:
+1

harai:
+1

analytically:
+1

elsmorian:
+1

CBarraford:
+1

cdesales:
+1

xpzouying:
+1

romainfun:
Note : Recent versions of logrotate include a maxsize option

nilstoedtmann:
logrotate is not watching logs continuously. By the time logrotate gets around to check for `[max]size`, the logs might have already filled up the disk (as mentioned in the OP).

fabrizioficci:
+1

sunkencity:
How about having an option to replacing the rotation scheme ending number with a postfix of date_from-date_to?

/my_process.log
/my_process.2016-04-13_04.30.00-2016-04-14_07.59.59.log
/my_process.2016-04-14_08.00.00-2016-04-14_12.00.00.log

Pros:
* easy to know in which logfile to look for older data.
* easy to add cron job to compress
* easy to add cron job to archive
* each logfile has a unique name
* still easy to delete the oldest file 
* no contention over logfiles being rotated while taking a backup
* can use supervisord without logrotate and thus not being prone to lose data while logs are rotated.

sunkencity:
If that is a way forward that is acceptable as a patch. I'd be happy to implement it.

keen99:
+1 - rather surprised (but the again, not considering issue with logfile ownership) that there's no option to compress rotated logs.

noxdafox:
An implementation effort was made long ago #619 but it was not good enough.

Main issue is that Supervisor is implemented on top of an event-loop and file compression might take long time making Supervisor missing important events.

EwaldvanGeffen:
[...]

PR #619, opened 2015-05-25
Automatic compression of backup log files

noxdafox:
This attemts to fix issue #322 

I included tests and documentation. Please let me know if anything else is required.

### DIFF
--- a/supervisor/loggers.py
+++ b/supervisor/loggers.py
@@ -159,7 +159,7 @@ class FileHandler(Handler):
 
 class RotatingFileHandler(FileHandler):
     def __init__(self, filename, mode='a', maxBytes=512*1024*1024,
-                 backupCount=10):
+                 backupCount=10, compress=False):
         """
         Open the specified file and use it as the stream for logging.
 
@@ -179,6 +179,8 @@ class RotatingFileHandler(FileHandler):
         respectively.
 
         If maxBytes is zero, rollover never occurs.
+
+        If compress is True, the backup files will be compressed as tar.gz.
         """
         if maxBytes > 0:
             mode = 'a' # doesn't make sense otherwise!
@@ -187,6 +189,7 @@ class RotatingFileHandler(FileHandler):
         self.backupCount = backupCount
         self.counter = 0
         self.every = 10
+        self.compress = compress
 
     def emit(self, record):
         """
@@ -238,13 +241,17 @@ class RotatingFileHandler(FileHandler):
 
         self.stream.close()
         if self.backupCount > 0:
+            strfmt = self.compress and "%s.%d.tar.gz" or "%s.%d"
+
             for i in range(self.backupCount - 1, 0, -1):
-                sfn = "%s.%d" % (self.baseFilename, i)
-                dfn = "%s.%d" % (self.baseFilename, i + 1)
+                sfn = strfmt % (self.baseFilename, i)
+                dfn = strfmt % (self.baseFilename, i + 1)
                 if os.path.exists(sfn):
                     self.removeAndRename(sfn, dfn)
             dfn = self.baseFilename + ".1"
             self.removeAndRename(self.baseFilename, dfn)
+            if self.compress:
+                handle_compression(dfn, "%s.%s" % (dfn, 'tar.gz'))
         self.stream = open(self.baseFilename, 'w')
 
 class LogRecord:
@@ -379,7 +386,8 @@ def handle_syslog(logger, fmt):
     handler.setLevel(logger.level)
     logger.addHandler(handler)
 
-def handle_file(logger, filename, fmt, rotating=False, maxbytes=0, backups=0):
+def handle_file(logger, filename, fmt, rotating=False, maxbytes=0, backups=0,
+                compress=False):
     if filename == 'syslog':
         handler = SyslogHandler()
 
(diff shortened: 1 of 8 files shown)


noxdafox:
Sorry for the extra commits but I don't have all Python versions on my dev env.

supervisor/loggers.py
```diff
@@ -387,10 +396,15 @@ def handle_file(logger, filename, fmt, rotating=False, maxbytes=0, backups=0):
...
 
     handler.setFormat(fmt)
     handler.setLevel(logger.level)
     logger.addHandler(handler)
 
     return logger
+
+def handle_compression(sfn, dfn):
+    with tarfile.open(dfn, "w:gz") as archive:
+        archive.add(sfn, arcname=os.path.basename(sfn))
```
mnaberez: This single operation can potentially block `supervisord` for many seconds, depending on the size of the log file being compressed, the system its running on, etc.  `supervisord` is a asynchronous program.  We can't block for long periods like this in `supervisord` because it will become unresponsive, and may destabilize programs running under it since it can't read their pipes if it's blocked.  

I don't see an easy solution to this given the way logging is currently implemented.  I think it may require compressing the log in pieces while returning control the event loop, or coming up with a way to run the compressor in an eventlistener subprocess.

noxdafox: Didn't consider that as I'm not 100% familiar with Supervisor's architecture.

Sure a subprocess could easily handle the the compression without blocking the daemon. I can provide such implementation if needed.

mnaberez: It may be possible to do this inside `supervisord` in a non-blocking way but I'm not sure how at the moment.  Doing it in an [eventlistener subprocess](http://supervisord.org/events.html#events) may be more straightforward and has no risk of blocking `supervisord` during compression.  If you haven't written an eventlistener before, there's some examples in the [Superlance](https://github.com/supervisor/superlance) package. 

An eventlistener can get info about all processes using the `getAllProcessInfo()` method of the XML-RPC interface, and this includes the path to the current log files of each process.

The eventlistener wouldn't have a way to know when log rotation occurred.  This could be addressed by adding a new event to `supervisord` to indicate when a rotation has occurred.  It may not be necessary to do this, the eventlistener could just glob every so often on a `TICK_` event.  If a `logfile.2` exists but no `logfile.1.gz`, time to gzip.

noxdafox:
I have been thinking about the proposed alternative with eventlistener subprocess and it seems a bit overengineered to me.

If we don't want to block supervisor with CPU bound jobs the easiest way to do so is with a separate Python process and a Queue. The daemon should simply take care of correctly terminate the process once stopped. Once we rotate, is enough to put the file path onto the queue and the dedicated process will take care of the compression.

To detect rare cases in which the amount of spammed log is so much the compression process cannot cope with, we simply have to poll the Queue state. If greater than, say 10 * logfile_backups, we can print warnings on the daemon log allowing the owner to spot the issue.

mnaberez:
> If we don't want to block supervisor with CPU bound jobs the easiest way to do so is with a separate Python process and a Queue. The daemon should simply take care of correctly terminate the process once stopped. 

There are many things that can go wrong with running a subprocess.  Making sure it terminates is one of them.  It also might not start, or might die unexpectedly while it's running, and all sorts of other things.  The eventlistener may seem "overengineered" but it allows us to handle all these conditions.  If we spawn subprocesses in another way, then we have a parallel system where we have to deal with all these issues again.  Supervisor users are accustomed to running eventlisteners since many are running things like (e.g. [memmon](https://github.com/Supervisor/superlance/blob/0.11/superlance/memmon.py)) already.

noxdafox:
This is indeed true, I'll proceed as agreed then.

mnaberez:
Closing for the reasons explained [above](https://github.com/Supervisor/supervisor/pull/619#discussion_r30990694).

noxdafox:
I started developing the event based plugin and I'm half way through but some questions came into my mind.

ATM I'm implementing it under superlance as a plugin but there are few things which I don't like.

The Supervisord is in charge of controlling the rotation. It would notify the happening with a "LOG_ROTATION" event with as extra parameter the path of the old logfile. The superlance plugin would react to it and compress the logfile.

Now the problem comes when it's time to delete the old logfiles. The Supervisor logger module is not aware (and it shouldn't) of any plugin. Yet the file has been renamed and its deletion would fail.

Possible solutions would be: 
 * do not rename the compressed files. I don't like it as it would be misleading for the end users
 * integrate the plugin within supervisor itself so the logger would be in control. In such case I need more information on how to correctly do that.

mnaberez:
> ATM I'm implementing it under superlance as a plugin but there are few things which I don't like.

Not all plugins need to be included in Superlance (most aren't).  We maintain a [list of plugins](http://supervisord.org/plugins.html) and there are other logging related plugins in that list.  New plugins are added to the list by making a pull request [list source](https://github.com/Supervisor/supervisor/blob/master/docs/plugins.rst).

noxdafox:
The problem is not whether to put the plugin in Superlance or in an indipendent project.

The problem is that as the code is within an external code-base, Supervisor cannot see that the logs it's rotating are changing name and format. Therefore deleting them will fail.

And, in my opinion, Supervisor should not behave differently according to which plugins are in use or not. That's one of the advantages of event driven architectures. It allows to de-couple the modules.

mnaberez:
I think the way people are handling this currently is to configure Supervisor with the regular non-rotating logger.  A program like logrotate is used and sends SIGUSR2 to Supervisor to signal it to reopen the logs.

PR #620, opened 2015-05-30
Exit with nonzero exit code if start fails

msabramo:
We've noticed that our deploy scripts declare success even if the process fails to start; this is because it seems that `supervisorctl start <app>` always returns an exit code of 0.

I only did this for `start` and didn't look at tests, but if it's agreed that this is a good idea in concept, then I can invest some more time in it.

Cc: @mnaberez, @djeebus, @georges

### DIFF
--- a/supervisor/supervisorctl.py
+++ b/supervisor/supervisorctl.py
@@ -700,8 +700,10 @@ class DefaultControllerPlugin(ControllerPluginBase):
         raise ValueError('Unknown result code %s for %s' % (code, name))
 
     def do_start(self, arg):
+        exitval = 0
+
         if not self.ctl.upcheck():
-            return
+            return exitval
 
         names = arg.split()
         supervisor = self.ctl.get_supervisor()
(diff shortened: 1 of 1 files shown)


msabramo:
and I suspected tests would need to be updated, but I'm going to wait to hear from @mnaberez about whether changing the API contract this way is a good idea before proceeding...

mnaberez:
I'm very much in favor of this, but I suggest if we do this that we do it for all commands and we document the expected exit codes for each command.  Similarly, I think it would be better if we send error messages to stderr (everything goes to stdout now).

This would be a breaking change so we probably do not want to release this in a Supervisor 3.x version.  Some scripts may depend on supervisorctl always returning 0, and a minor version release of Supervisor shouldn't break those scripts.  We could include it in Supervisor 4 (master branch).

mattwiller:
This seems pretty similar to #292 -- I would LOVE to see this added, since it's causing us a lot of pain not being able to easily check if an operation succeeded or not.

mnaberez:
Thanks for the patch but there's now a newer one in #668 that adds exit codes for all errors.  I intend to merge #668 or some version of it, so I'm going to close this PR in favor of that one.  I'd appreciate if you could please check out that PR and leave comments on it with your thoughts.

PR #625, opened 2015-06-14
Added optional waiter parameter to listener.wait()

vovanec:
Added optional waiter parameter to EventListenerProtocol.wait() method. Waiter is typically threading.Event() instance which allows handle signals more gracefully for eventlistener applications i.e. allow application to complete check cycle before exiting.

Example usage in my eventlistener process health check application:

```python
    def run(self):

        self._install_signal_handlers()

        while not self._stop_event.is_set():
            try:
                headers, payload = childutils.listener.wait(sys.stdin, sys.stdout, waiter=self._stop_event)
            except childutils.WaitInterrupted:
                self._log('Health check for %s process group has been told to stop.', self._process_group)

                break
            
            ... Do custom checks and restart programs if needed.

            childutils.listener.ok(sys.stdout)

        self._log('Done.')
```

### DIFF
--- a/supervisor/childutils.py
+++ b/supervisor/childutils.py
@@ -1,3 +1,4 @@
+import select
 import sys
 import time
 
@@ -49,10 +50,18 @@ class ProcessCommunicationsProtocol:
 
 pcomm = ProcessCommunicationsProtocol()
 
+
+class WaitInterrupted(Exception):
+    pass
+
+
 class EventListenerProtocol:
-    def wait(self, stdin=sys.stdin, stdout=sys.stdout):
+    
+    SELECT_TIMEOUT = 0.5
+    
+    def wait(self, stdin=sys.stdin, stdout=sys.stdout, waiter=None):
         self.ready(stdout)
-        line = stdin.readline()
+        line = self._readline(stdin, waiter=waiter)
         headers = get_headers(line)
         payload = stdin.read(int(headers['len']))
         return headers, payload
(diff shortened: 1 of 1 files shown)


vovanec:
Here's the use case: https://github.com/vovanec/supervisor_checks/blob/master/supervisor_checks/check_runner.py#L65

mnaberez:
Thanks for your work on this pull request but I have decided not to apply it. 

I wouldn't generally have issues adding an optional argument to add a new feature, but I think this case is a little different.  `childutils` is an API that has been around since at least 2008 unchanged. I don't think it's had any issues reported during that time.  Eventlisteners using `childutils` will work with any Supervisor version since we added the feature.  This is easy for users.

If we make this proposed addition, there will now by the possibility of eventlisteners being written that require specific versions of `childutils`, so they will only work on some versions of Supervisor.  I think this should be avoided if possible to keep things easy for users installing plugins.  I'm not saying it has to be frozen forever, but I think there should be a really compelling reason to introduce sources of plugin incompatibilities.  I think your use case is valid but I don't think it's quite worth the compatibility issue.

I'm going to ask that you please include this in your own code.  I think this will be better because it avoids the issues above and also because it gives you a much larger audience: your eventlisteners will work on all past Supervisor versions.

vovanec:
okay, it makes sense. I will make this the part of check framework.

Thanks
--Vovan

PR #626, opened 2015-06-24
pidproxy: support for daemons which terminate immediately and do not provide a no-detach option

neta79:
Daemon which do not expose a 'foreground' feature can now be handled via the
--wait parameter. Pidproxy will wait the specified amount of seconds for valid
pidfile to appear, then will monitor the pid in order to figure out if the proxied
process has died.
The pidfile must refer to an existing pid in order to be considered valid.

### DIFF
--- a/supervisor/pidproxy.py
+++ b/supervisor/pidproxy.py
@@ -21,7 +21,6 @@ class PidProxy:
         self.cmdargs = cmdargs
 
     def go(self):
-        print(self.cmdargs)
         self.pid = os.spawnv(os.P_NOWAIT, self.command, self.cmdargs)
         if self.wait_pidfile_time:
             self.wait_pidfile_showup()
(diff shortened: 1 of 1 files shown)


neta79:
Hi. I needed some pidproxy support for daemons which just fork into background without providing a no-detach option.
This is my attempt at solving the problem. It basically augments pidproxy with the option (--daemon) to wait for a pidfile to appear (with a timeout) and then monitor its pid for validity.
It also introduces argparse; this would potentially conflict with other pull requests which implement custom-written argument parsing (but adds the recommended stuff to deal with them).

jasonrhaas:
@mnaberez Are you planning on merging in this pull request?  I found that the `pidproxy.py` program that comes with supervisor did not work properly for "running a daemon process in supervisor".  I was looking for a solution and found that this pull request does the job nicely :+1:

supervisor/pidproxy.py
```diff
@@ -7,22 +7,27 @@
 import sys
 import signal
 import time
+import argparse
```
mnaberez: I haven't really reviewed the rest of this pull request but Supervisor works on Python 2.6+ so it can't use `argparse`.

neta79:
Other than the argparse issue, I have found later that there are side cases when this patch does not work nicely.  Primarily, when proxying process trees which tend to respawn: it may just leave orphan processes around.
I'd recommend **against** pulling it, in the current status. It needs better process tree handling.

PR #630, opened 2015-07-06
Make readFile filter control chars 0x1F and below, except for TAB, CR and LF.

jmtapio:
This is needed because there might be control characters like ESC in
log files and xmlrpclib marshaller does not process them leading into
invalid XML.

You can typically run into this when processes output ansi colors. The esc sequences are passed as is and typically cause a crash in a xml-rpc client connecting to supervisord and requesting log files stdout or stderr. The control chars need to be either escaped or removed. Given this method is used for log files I propose filtering them.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1956,7 +1956,11 @@ class FastCGIGroupConfig(ProcessGroupConfig):
 
 def readFile(filename, offset, length):
     """ Read length bytes from the file named by filename starting at
-    offset """
+    offset
+
+    Filter control characters x0-x1f except for tab, cr and lf. They would
+    cause problems when marshalling later on.
+    """
 
     absoffset = abs(offset)
     abslength = abs(length)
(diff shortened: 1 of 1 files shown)


mnaberez:
> You can typically run into this when processes output ansi colors.

The following can be set in the `[supervisord]` section of the config file ([reference](https://github.com/Supervisor/supervisor/blob/3.0/supervisor/skel/sample.conf#L37)):

    [supervisord]
    strip_ansi=true            ; (strip ansi escape codes in logs; def. false)

`strip_ansi` has the advantage of filtering out the entire control code.  This patch would only filter out the lead-in character (ESC), so it would show control code artifacts, e.g. `[2JHello world`.

jmtapio:
Thanks, that lets me bypass the problem in my current case but it still does lets some control chars through breaking the xml. Kinda surprising that the Python xmlrpc dumps does not do escaping for them.

mnaberez:
> Thanks, that lets me bypass the problem in my current case but it still does lets some control chars through breaking the xml.

What control codes are you seeing that does it not catch?

jmtapio:
On Thu, Jul 09, 2015 at 09:46:59AM -0700, Mike Naberezny wrote:
> > Thanks, that lets me bypass the problem in my current case but it still does lets some control chars through breaking the xml.
> 
> What control codes are you seeing that does it not catch?

NUL / \x0 goes through crashing xml-rpc client for example. Basically
the current state of filtering requires that all the programs run by
supervisor filter non-ansi control stuff from their output.

Granted, control chars in output other than ansi sequences do not
happen very often.

PR #633, opened 2015-07-16
Wrap exceptions and errors(interactive mode feedback) with SystemExit

lukeweber:
Related: https://github.com/Supervisor/supervisor/issues/24

Problem: People using supervisor inside of bash/terminal for a single command don't receive exit codes consistently for various errors, and erroneously get exit with code 0, which in LSB terms means success. The alternative is to parse the command output from supervisorctl, which is error prone and difficult in bash.

- [x] implement handle_error(message=None, fatal=False)
- [x]  Wrap any output that contains "Error" or "ERROR" in supervisorctl.py with a handle_error(message=...) method.
- [x]  Wrap any blank "raise" with handle_error(fatal=True)
- [x]  Add an option exit_on_error in realize() in ClientOptions defined as !interactive with default of 0. I've chosen not to use !interactive directly in supervisorctl in case someone extends this directly and does want internal exceptions, but not interactive mode.
- [x]  Create duplicate tests for all tests that check for different Exceptions in test_supervisorctl.py to check for SystemExit when options.exit_on_error = True

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -318,7 +318,7 @@ follows.
 ``nocleanup``
 
   Prevent supervisord from clearing any existing ``AUTO``
-  chlild log files at startup time.  Useful for debugging.
+  child log files at startup time.  Useful for debugging.
 
   *Default*:  false
 


lukeweber:
I got tests to finally pass at the expense of hard setting mock==1.0.1 in tox.ini.  Current version mock==1.1.3 doesn't play nice with python 2.6. I tried conditional building for python 2.6, but it's not clear if that's possible in tox.ini and I couldn't get it to work in a requirements file either. I also found these guys having trouble with the same issue:
https://bugs.launchpad.net/python-manilaclient/+bug/1474261

I ended up having two handle_error methods, one in the controller and  another in plugin(copied), because I couldn't reference handle_error in the controller in tests because DummyController doesn't extend controller. DummyController could extend Controller and override most behavior, but keep my handle_error untouched, or something else.

Issue #250, opened 2013-06-25
Add supervisorctl commands to documentation


grillermo:
There is a nice intro on how to run it on the docs, and configs and everything, but nowhere it says what it does, what it is, and why do you want it.
Or maybe i skipped some part of the docs.

oryband:
:+1: it is terrible that the docs are missing any `supervisorctl` documentation
please add

mnaberez:
@oryband Pull request welcome.

VynceMontgomery:
@mnaberez of course, the reason people find this issue is that they don't know what it does. Bit of a chicken-and-egg problem, y'see. ("Pull requests welcome" makes a lot less sense as a response to requests for documentation.)

mnaberez:
@VynceMontgomery At a `supervisorctl` prompt, `help` lists all commands and `help <command>` shows specific help for a command.  All of the commands are implemented in one file, [supervisorctl](https://github.com/Supervisor/supervisor/blob/3.1.3/supervisor/supervisorctl.py).  We'd certainly appreciate a pull request if someone is up to spending a little time on it, hopefully this provides a starting point.

Somebi:
Is this fixed? Bump if not. Very frustrating.

mnaberez:
Fixed in #866.

PR #641, opened 2015-08-04
Add exit codes for error cases

lukeweber:
- do_status -> Return exit with code=7 in one matching process is in stopped state
- exit with non-zero codes in the event of errors while running non-interactive

Fixes https://github.com/Supervisor/supervisor/issues/24

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -318,7 +318,7 @@ follows.
 ``nocleanup``
 
   Prevent supervisord from clearing any existing ``AUTO``
-  chlild log files at startup time.  Useful for debugging.
+  child log files at startup time.  Useful for debugging.
 
   *Default*:  false
 


jmagnusson:
Thumbs up for this if it works as advertised! :+1:

lukeweber:
I'll manually run through the cases today, and give an update.

lukeweber:
Latest commit, I added a context manager to handle errors that may occur in a loop, so we can print them out, and exit with an error when we exit the loop. Also I added handle_xmlrpc_fault_state which triggers an error for any faults and aren't fault.SUCCESS, otherwise just prints them as before using the _(start|stop|clear)result method.

Multi command mode will run through the commands, print the errors on the fly, and then return with the code of the first error that it hit.

If anyone else wanted to run through some test scenarios and see if anyone notices anything would be cool and appreciated.

```
> ./bin/supervisorctl -c /etc/supervisor/supervisor.conf pid backend test
0
No such process test
> echo $?
1
```

```
> ./bin/supervisorctl -c /etc/supervisor/supervisor.conf status backend test
test: ERROR (no such process)
backend                FATAL     Exited too quickly (process log may have details)
> echo $?
5
```

Single commands work as expected: 
```
> ./bin/supervisorctl -c /etc/supervisor/supervisor.conf status test
test: ERROR (no such process)
> echo $?
5
```

```
> ./bin/supervisorctl -c /etc/supervisor/supervisor.conf status backend
backend                FATAL     Exited too quickly (process log may have details)
> echo $?
7
```

lukeweber:
Moving over a comment from https://github.com/Supervisor/supervisor/issues/24

@gdm85 wrote: what about adding some script-based integration tests? I can see that a simple command like sh -c 'sleep 3 && true' and sh -c 'sleep 3 && false' could do the job

====
Would be nice to have some feedback from the project owners as to how they would like to see something like this implemented. You could create a set of tests and run a supervisord like they do in EntryPointTests in test_supervisord.py and then run supervisorctl against it with exec. Not even sure if that would work as I'm not certain if the process runs while it's in scope or just a single run on that line. Maybe I would create a context manager for testing i.e. "with SupervisorDaemon:" that starts the server in my test and lets me run the supervisorctl on the command line against it.

mnaberez:
> Would be nice to have some feedback from the project owners as to how they would like to see something like this implemented.

I suggested something like this to the Supervisor author some years ago but he thought at the time that the unit tests were sufficient.

lukeweber:
> You may not want to spend time on it unless his position has changed.

@mcdonc - Do we want to stick with unit tests only or is this something you feel like some partial integration test might be in order?

gdm85:
For the records, I proposed it because it feels like a natural type of tests to add for a **supervisor** and could provide tests that ensure deterministic behavior under all program action exit paths we want to put a brace on.

mnaberez:
Thanks for this pull request.  I haven't reviewed this implementation yet, but a big issue I have approaching this is that the `supervisorctl` commands are not documented (#250).  This is a problem because I think the exit statuses each command will now produce must be documented when we do this.  I would also really like us to unit test the exit statuses for each command because they will become an API contract once adopted and we have to be careful not to break them.  I'm definitely willing to help with these tasks (#250 needs to get done regardless) but I am currently working on other bugs and I'm not sure when I'll get to it.

lukeweber:
That seems fine. I think I would then say that integration testing would
feel the most natural. Any ideas on how to approach that or if we can?

mnaberez:
Most of the `supervisorctl` commands have unit test coverage including the error conditions.  This is the way I would want to test it:

    # this is existing test in test_supervisorctl.py
    def test_start_badname(self):
        plugin = self._makeOne()
        result = plugin.do_start('BAD_NAME')
        self.assertEqual(result, None)
        self.assertEqual(plugin.ctl.stdout.getvalue(),
                         'BAD_NAME: ERROR (no such process)\n')
        self.assertEqual(plugin.ctl.exitstatus, 1) # something like this could be added

If the new code can't be unit tested easily like the existing code, that may be a signal that it should be refactored somehow.

lukeweber:
Good point. Thanks for the helpful feedback. I cleaned things up.

Pending
- [ ] onecmd wrapper review - I had to create a onecmd_run method with the old contents of onecmd so I could effectively throw an exception for all the ways it could possibly exit. 
- [ ] More testing
- [ ] Review exit codes. do_status and the rest might be distinct, although non-zero is the most important.

mnaberez:
Unanswered question from #24:
> Is supervisorctl.py code used for anything besides single commands or interactive mode(two states), or is it reused in libraries (third state - not interactive, raise exceptions and don't exit on errors)?

These are the cases I can think of when `supervisorctl` can run multiple commands:

- `supervisorctl` commands can be combined with `;` ([code](https://github.com/Supervisor/supervisor/blob/3.1.3/supervisor/supervisorctl.py#L127)).
- Some commands call other commands, e.g. `restart` is `stop`+`start` ([code](https://github.com/Supervisor/supervisor/blob/3.1.3/supervisor/supervisorctl.py#L800-L801)).
- Plugins.  They can potentially call any `do_` method like `restart` does above.  

`supervisorctl` is built on [cmd](https://docs.python.org/2/library/cmd.html).  Normally the `do_` methods that `cmd` calls always return and don't raise.  We should probably try to keep doing that for compatibility with the cases above (particularly plugins, there are many and we don't know what they all do).

lukeweber:
Ok, I'm feeling fairly good about things generally. I don't think I violate the current apis except that if someone uses supervisorctl directly in single command mode, it will have relevant exit codes. The overview is as follows. 

- onecmd is the single entry point for running a command from supervisorctl.py, so that's also the primary point where SystemExit(code) would be raised if eror_status is set and exit_on_error is set.
- exit_on_error is only activated in one condition inside options with the following code:
```python
self.exit_on_error = 0 if self.interactive else 1
```
- handle_error now wraps any string in supervisorctl that previously had an error.
- do_status and process status is special so we set code=4 for unknown process names, and 3 for processes in stopped state.
- I wrap upcheck  method in status to also make sure it also sets code = 4 for any of it's errors. ("program or service status is unknown")
- I stuck generally with code=1 for all general errors. ("generic or unspecified error (current practice)")
- When upcheck has an error outside of do_status it sets four different codes error_codes, but I'll let everyone just have a look.

@mnaberez  - Should we create LSBErrorCodes, and LSBStatusErrorCodes enums to make this clearer as a final step? naming?

lukeweber:
@mnaberez - Thanks for all the feedback.I just did a final review of codes and reread the [LSB spec](http://refspecs.linuxfoundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/iniscrptact.html) and I think we're ready for final review from your side.

I've removed some error cases, specifically in case start or stop where things are already in a started or stopped or stopping state as those should not trigger an error.

I also added constants for LSBInitErrorCode and LSBStatusErrorCode to do away with the magic so it's clearer which exit code we're setting.

lukeweber:
@mnaberez - As far as I know this patch is done. Let me know if there's anything else that should be done with this one.

lukeweber:
Anyone ready to merge this?

jmagnusson:
Seems like they are busy doing something else.

lukeweber:
@mnaberez - Sorry this one got closed. Wanted to change my master so I could start doing some dev on this one myself in my fork. I've squashed my merges though and added documentation. Maybe you just want to merge my master though and get ahead of the game It's only 7 commits and a lot of awesome including full tests and documentation......

https://github.com/lukeweber/supervisor/commits/master

Issue #282, opened 2013-08-19
supervisor creates useless empty stderr log files with redirect_stderr=True


timabbott:
With redirect_stderr=True, supervisord will still end up creating automatic mktemp()'d files under /var/log/supervisor for the stderr log file, even though nothing is getting written to it.

coderlol:
+1

PR #643, opened 2015-08-11
Empty stderr logs no longer created when redirect_stderr=True. Closes…

evalapply:
… #282


evalapply:
When I run the tests with nosetests they all pass :( I will just replace assertIs with assertEqual unless it's the CI setup that should be changed (?).

PR #508, opened 2014-10-27
Fix #369 relative command path

schemacs:
The docs say:

```
The command can be either absolute (e.g. /path/to/programname) or relative (e.g. programname). 
If it is relative, the supervisord’s environment $PATH will be searched for the executable.
```

This fixes relative path like to/programname as stated in #369 .

PR #648, opened 2015-08-19
Fix #369 relative command path (New)

### LINKED ISSUE #369: Directory not searched for executable
**EDIT**: Re-parsing the docs I see this probably counts as "expected behavior" however it doesn't seem very obvious or clear, so perhaps this is more of a feature request.

---

I think this is because `process.get_execv_args` only checks my path for the executable and not the directory set by `directory`.

**example**

```
mal:~# ls -l /opt/hello/bin/hello
-rwxrwxr-x 1 mal mal 1088 Jan 29 14:27 /opt/hello/bin/hello
mal:~# sudo supervisorctl status
example          BACKOFF    can't find command 'bin/hello'
```

```
[program:example]
command = bin/hello
directory = /opt/hello
```

lukeweber:
1. Uses isabs() for program check which was mentioned here as a desired fix and makes sense.
1. Inserts program directory into position 0 in path search for program, which is similar to how path search works in other programs when executing a command from a directory.
1. Now has two tests for having the program in a custom path, and for having the program in the program directory.

Incorporates feedback from original pull:
https://github.com/Supervisor/supervisor/pull/508

Resolves:
https://github.com/Supervisor/supervisor/issues/369

I would update the docs.  "If it is relative, the supervisord’s environment $PATH will be searched for the executable." => "If it is relative, the program's directory and supervisord’s environment $PATH will be searched for the executable."

### DIFF
--- a/supervisor/process.py
+++ b/supervisor/process.py
@@ -115,19 +115,18 @@ class Subprocess(object):
         else:
             raise BadCommand("command is empty")
 
-        if "/" in program:
-            if program.startswith('/'):
-                filename = program
-            else:
-                filename = os.path.join(self.config.directory, program)
-
+        if os.path.isabs(program):
+            filename = program
             try:
                 st = self.config.options.stat(filename)
             except OSError:
                 st = None
-
         else:
             path = self.config.options.get_path()
+            # Search the root config directory of this program if it's set and is absolute.
+            if self.config.directory is not None and os.path.isabs(self.config.directory):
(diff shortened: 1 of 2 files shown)


mnaberez:
This is so close to #508 that I am going to close it as a duplicate.

lukeweber:
@mnaberez - As far as I know this patch is done.

laoyur:
still no luck to get this merged after 2 years ?

PR #651, opened 2015-08-20
Improve Medusa zombie connection detection

### LINKED ISSUE #650: XMLRPC prevents persistent connections (Broken pipe)
The HTTP-Server checks every 500 connections made for connections older than 30 minutes and closes them (https://github.com/Supervisor/supervisor/blob/master/supervisor/medusa/http_server.py#L505) This means it's not possible to keep a persistent connection open as it will be eventually closed after enough newer ones came in.
I think it would be better to check a last used timestamp instead of the creation time

sfriesel:
Medusa checked every 500 connections made for connections older
than 30 minutes and closed them. This made it impossible to keep
a persistent connection when many short-lived connections happen at
the same time. It will now only close connections that have been
inactive for that amount of time, i.e. no send/recv activity.

Fixes #650

### DIFF
--- a/supervisor/medusa/CHANGES.txt
+++ b/supervisor/medusa/CHANGES.txt
@@ -4,6 +4,7 @@ PATCHES MADE ONLY TO THIS MEDUSA PACKAGE BUNDLED WITH SUPERVISOR
   of these modules introduced backward-incompatible changes.
 * Changed imports throughout from "medusa" to "supervisor.medusa".
 * Removed medusa files not used by Supervisor.
+* Time out connections based on inactivity instead of age
 
 Version 0.5.5:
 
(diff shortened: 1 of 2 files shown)


sfriesel:
This is my first naive attempt. I don't know if `creation_time` is a public attribute `http_channel`, in that case I could put it back in.

mnaberez:
> This is my first naive attempt. I don't know if creation_time is a public attribute http_channel, in that case I could put it back in.

I don't know either. Medusa doesn't have much in the way of tests or docs to help us.  Let's just leave it alone in case any of this ever gets merged with Medusa upstream (unlikely, but who knows).  Please add a note about this change to the [Medusa changelog](https://github.com/Supervisor/supervisor/blob/master/supervisor/medusa/CHANGES.txt).

sfriesel:
@mnaberez updated with your suggestions and squashed

mnaberez:
Merged. Thanks!

Issue #640, opened 2015-08-04
Missing trailing slash on inet_http_server redirects


rafis:
Then inet_http_server do some redirect it missing trailing slash 

    http://127.0.0.1:9211?message=Process%20progam1%20started

should be

    http://127.0.0.1:9211/?message=Process%20progam1%20started

I tried use nginx configuration file with `proxy_pass default`, but it not works because of missing trailing slash:

    location / {
        proxy_pass http://127.0.0.1:9211;
        proxy_redirect default;
    }

I have to use following fix:

    location / {
        proxy_pass http://127.0.0.1:9211;
        proxy_redirect default;
        proxy_redirect http://127.0.0.1:9211 $scheme://$host/; # fix for supervisord inet_http_server
    }

roadrunner:
it has been passed more than 2 years when first reported.
a lot bug report and PRs.. 
thanks for not solving this shitty problem.

chavenor:
Wow just saved me.   Again I'm confused to as why this hasn't been merged back in?

PR #593, opened 2015-03-19
Url prefix for web interface

calvin:
This is a pull request to resolve issue #29 and #372.
Instead of relying on `HTTP_REFERER` like in #28, I have introduced a new option.

This PR also addresses issue #495.

PR #669, opened 2015-10-07
Fix for Missing trailing slash on inet_http_server redirects - #640

lukeweber:
get_server_url() strips the trailing "/", so it has to be added back on the redirect between the host and http args.

http.py
```python
    def get_server_url(self):
        """ Functionality that medusa's http request doesn't have; set an
        attribute named 'server_url' on the request based on the Host: header
        """
        default_port={'http': '80', 'https': '443'}
        environ = self.cgi_environment()
        if (environ.get('HTTPS') in ('on', 'ON') or
            environ.get('SERVER_PORT_SECURE') == "1"):
            # XXX this will currently never be true
            protocol = 'https'
        else:
            protocol = 'http'

        if 'HTTP_HOST' in environ:
            host = environ['HTTP_HOST'].strip()
            hostname, port = urllib.splitport(host)
        else:
            hostname = environ['SERVER_NAME'].strip()
            port = environ['SERVER_PORT']

        if port is None or default_port[protocol] == port:
            host = hostname
        else:
            host = hostname + ':' + port
        server_url = '%s://%s' % (protocol, host)
        if server_url[-1:]=='/':
            server_url=server_url[:-1]
        return server_url
```

### DIFF
--- a/supervisor/web.py
+++ b/supervisor/web.py
@@ -427,7 +427,7 @@ class StatusView(MeldView):
                     return NOT_DONE_YET
                 if message is not None:
                     server_url = form['SERVER_URL']
-                    location = server_url + '?message=%s' % urllib.quote(
+                    location = server_url + "/" + '?message=%s' % urllib.quote(
                         message)
                     response['headers']['Location'] = location
 


mnaberez:
@rafis @lukeweber @calvin Could you guys please review this against #593?  

That pull request is another attempt at better proxy support that also makes the URL more configurable.  If #593 also fixes this issue maybe we should just merge that.  I'm not using the web interface behind a proxy server so I appreciate your feedback.

calvin:
#593 fixes this by introducing new `base_path` setting. Also it ensures the trailing `/` always exists. (https://github.com/Supervisor/supervisor/pull/593/files#diff-0240b66c1a20db854f3d6eacb71caa32R1031)

PR #670, opened 2015-10-07
Expose more info from getAllConfigInfo

msabramo:
This lets clients know more about what the current settings are that supervisord is using, which might be different from what's in the config files on disk, if the files have changed recently and a `supervisorctl update` hasn't been done yet.

See https://github.com/Supervisor/supervisor/pull/672 for a sample usage of this feature (a new `supervisorctl` command).


msabramo:
I'll try to fix tests

msabramo:
OK, tests fixed.

mnaberez:
Thanks!  

The `getAllConfigInfo()` RPC method was added with the config reloading patch submitted by Anders Quist.  I believe that it is only used by the `avail` command in `supervisorctl`, also from that patch.  Your additions won't affect the output of `avail` as far as I can tell, and I can't think of any good reason not to expose this info, so I've merged it.

mikeyg123:
Can this method be documented on the website http://supervisord.org/api.html#process-control

PR #673, opened 2015-10-11
Don't recurse when reaping

aidanhs:
Stumbled across this code, thought it could be simplified.

Removes an unnecessary test as well.

### DIFF
--- a/supervisor/supervisord.py
+++ b/supervisor/supervisord.py
@@ -275,21 +275,18 @@ class Supervisor:
                 self.ticks[period] = this_tick
                 events.notify(event(this_tick, self))
 
-    def reap(self, once=False, recursionguard=0):
-        if recursionguard == 100:
-            return
-        pid, sts = self.options.waitpid()
-        if pid:
+    def reap(self, once=False):
+        maxreap = 1 if once else 100
+        for i in range(maxreap):
+            pid, sts = self.options.waitpid()
+            if not pid:
+                return
             process = self.options.pidhistory.get(pid, None)
             if process is None:
                 self.options.logger.info('reaped unknown pid %s' % pid)
             else:
                 process.finish(pid, sts)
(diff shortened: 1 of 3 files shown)


mnaberez:
Thanks, but I don't see any reason to change this.

aidanhs:
I found the code confusing when I initially came across it. This change seemed like a decent clarity improvement (fewer function arguments, fewer lines of code, less complexity to need to test) but it is of course your call.

Have you considered a contribution guide to indicate what kind of thing you're interested in? Something like:

> Pull requests for bug fixes and features are welcome, though significant features should be discussed on the mailing list beforehand.

> General refactoring will not be merged unless there is an extremely compelling reason to do so.

I can make a PR to add this if you agree with the general wording? It'll help prevent wasted time on PRs like this one.

mnaberez:
> General refactoring will not be merged unless there is an extremely compelling reason to do so.

I don't agree with this wording and I also don't see a reason to add such disclaimers.  Many patches are merged including refactoring.  They are reviewed on a case by case as they are submitted.  I understand that you think this is an improvement, I just don't happen to agree with you on this, sorry.

aidanhs:
Fair enough.

aidanhs:
To elaborate a bit more as you gave such a comprehensive response, which I appreciate - you make a very good point regarding external calls which may break, I hadn't thought about that.

I definitely understand the problems of trying to keep backwards compatibility in mature programs, so this is very reasonable.

> I probably should have given you more of a response why, but you didn't offer much with the patch either.

You're right, I apologise for that.

mnaberez:
Thank you for the patch, I do appreciate it, and please feel free to submit more patches in the future.  It's an old program so there's some issues around that but in general we do want to improve it.

aidanhs:
Willdo! :)

PR #676, opened 2015-10-22
Write activity log to syslog when logfile=syslog

icy:
When main configuration logfile=syslog, ServerOptions will not call (normalize_path)
to convert that option value to a real path (i.e., working_dir/syslog). Instead,
it passes that option to future loggers and that will make actitivy logs come to
(local) syslog receiver.


icy:
Possibly better PR in 162

PR #684, opened 2015-11-04
Add authentication prompts for non-interactive mode

jasonzzz:
Supervisor only support authentication prompts for interactive mode, like:

```
$ supervisorctl -c /etc/supervisor/supervisord.conf
Server requires authentication
Username:
Password:
```

For those who want to use command args:

```
$ supervisorctl -c /etc/supervisor/supervisord.conf status
Error: Server requires authentication
For help, use /usr/local/bin/supervisorctl -h
```

Here comment out the `if` condition to add prompts instead of just printing out the usage for non-interactive mode.

### DIFF
--- a/supervisor/supervisorctl.py
+++ b/supervisor/supervisorctl.py
@@ -183,16 +183,13 @@ class Controller(cmd.Cmd):
                     return do_func(arg)
                 except xmlrpclib.ProtocolError as e:
                     if e.errcode == 401:
-                        if self.options.interactive:
-                            self.output('Server requires authentication')
-                            username = raw_input('Username:')
-                            password = getpass.getpass(prompt='Password:')
-                            self.output('')
-                            self.options.username = username
-                            self.options.password = password
-                            return self.onecmd(origline)
-                        else:
-                            self.options.usage('Server requires authentication')
+                        self.output('Server requires authentication')
+                        username = raw_input('Username:')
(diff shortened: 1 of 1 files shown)


mnaberez:
This can't be merged because a large number of deployments run `supervisorctl` commands through some sort of automation tool.  For all those systems, if authentication fails for any reason, `supervisorctl` would hang indefinitely on a prompt with nobody to answer it.

jasonzzz:
> if authentication fails for any reason, supervisorctl would hang indefinitely on a prompt with nobody to answer it.

Could we accept exiting `supervisorctl` with `ECONNREFUSED` instead of trying again with prompts if the authentication fails?

Also I'm wondering if we could support basic http authentication method in the `serverurl` option, like `supervisorctl --serverurl 'http://user:pass@host:port'`?

mnaberez:
> Could we accept exiting supervisorctl with ECONNREFUSED instead of trying again with prompts if the authentication fails in this case?

I don't think we should be the possibility of `supervisorctl` hanging at all in noninteractive mode, so I don't want to merge this or a variation of it.

> Also for those automation tools, I'm wondering if we could support basic http authentication method in the serverurl option, like supervisorctl --serverurl 'http://user:pass@host:port'?

The `--username` and `--password` options do this.  See `supervisorctl --help`.

jasonzzz:
Forgotten `--username` and `--password`. Thanks.

PR #692, opened 2015-11-14
Test that doc building + readme are correct

Ivoz:
Currently you can see at [supervisor's pypi page](https://pypi.python.org/pypi/supervisor) that the long_description has failed to render.

This additional test env will test that
A) the docs build correctly
B) The readme can be rendered correctly

Very useful for making sure your readme changes don't bork anything (like pypi rendering).

the [readme](https://pypi.python.org/pypi/readme) package integrates into the `setup.py check -r -m -s` call.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    cover,cover3,py26,py27,py32,py33,py34,py35,pypy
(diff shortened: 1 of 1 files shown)


mnaberez:
Committed in 8911249648525129403f588caa1a686d9f29780c, thanks.

PR #737, opened 2016-04-11
expand the host node name option on includes

plockaby:
This adds support for this syntax in an include:

```
[include]
files = /etc/supervisor/supervisor.d/%(host_node_name)s/*.conf
```

This allows me to deploy the same directories to all of my hosts and let supervisor figure out which ones to load by what host it is running on.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -570,7 +570,9 @@ class ServerOptions(Options):
             if need_close:
                 fp.close()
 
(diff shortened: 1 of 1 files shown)


plockaby:
I had previously submitted #293 which wasn't accepted in favor of commit 2d6ca34 but that commit did not enable this specific functionality.

Issue #122, opened 2012-05-30
supervisord Starts All Processes at the Same Time


fredpalmer:
My config:

```
[program:redis-testapp]
command=/opt/bcs/bin/redis-server /apps/testapp/releases/current/environments/all/redis.conf
stdout_logfile=/var/log/redis_testapp_log
stderr_logfile=/var/log/redis_testapp_log
startsecs=30
priority=1
autostart=true
autorestart=true

[program:celerybeat-testapp]
command=python -O manage.py celerybeat --loglevel=INFO --schedule=/apps/testapp/db/celerybeat_schedule_db
stdout_logfile=/var/log/celerybeat_testapp_log
stderr_logfile=/var/log/celerybeat_testapp_log
priority=999
startsecs=5
autostart=true

[program:celery-testapp]
command=python -O manage.py celeryd --loglevel=INFO --events
stdout_logfile=/var/log/celeryd_testapp_log
stderr_logfile=/var/log/celeryd_testapp_log
priority=100
startsecs=10
autostart=true

[program:gunicorn-testapp]
command=gunicorn_django --workers=10 --log-level info --timeout 500 --bind=127.0.0.1:8004
stdout_logfile=/var/log/gunicorn_testapp_log
stderr_logfile=/var/log/gunicorn_testapp_log
priority=999
startsecs=10
autostart=true

[program:memcached-testapp]
command=/opt/bcs/bin/memcached -m 128 -l 127.0.0.1 -p 11212 -u nobody -P /apps/testapp/run/memcached.pid
stdout_logfile=/var/log/memcached_testapp_log
stderr_logfile=/var/log/memcached_testapp_log
priority=11
autostart=true
autorestart=true


```

My output =>

```
2012-05-30 22:37:33,181 INFO daemonizing the supervisord process
2012-05-30 22:37:33,182 INFO supervisord started with pid 16230
2012-05-30 22:37:34,195 INFO spawned: 'redis-testapp' with pid 16232
2012-05-30 22:37:34,206 INFO spawned: 'memcached-testapp' with pid 16233
2012-05-30 22:37:34,214 INFO spawned: 'celery-testapp' with pid 16234
2012-05-30 22:37:34,238 INFO spawned: 'celerybeat-testapp' with pid 16235
2012-05-30 22:37:34,477 INFO spawned: 'gunicorn-testapp' with pid 16241
[...]

PR #747, opened 2016-04-24
Add support for Docker within supervisor

maihde:
I like using supervisor to manage my Docker processes, but it doing so has one major downside, there are many situations where a docker container will stop but will not be removed.  When this happens supervisor is unable to restart the process because the container name is already taken (because I prefer to have my container names match the supervisor process names).

These changes allow supervisor to be fully aware of managing Docker containers as part of the regular process management.  I had considered making this a plugin, but the rpcinterface plugin mechanism didn't allow me to easily extend the configuration file syntax for the [program] sections.

The only significant design decision that I made was using subprocess to issue the docker stop/rm commands.  This means that supervisor would require at least Python 2.4.  If that is a problem I can replace it with alternate approaches.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -38,6 +38,9 @@
   number of tickets as a source of confusion and upgrade difficulties, so the
   path has been added.  Patch by Kelvin Wong.
 
+- Glob patterns in the ``include`` section now support the ``host_node_name``
+  expansion.  Patch by Paul Lockaby.
+
 - Files included via the ``[include]`` section are now logged at the ``INFO``
   level instead of ``WARN``.  Patch by Daniel Hahler.
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1119,14 +1119,18 @@ configuration.
   includes it.  A "glob" is a file pattern which matches a specified
   pattern according to the rules used by the Unix shell. No tilde
   expansion is done, but ``*``, ``?``, and character ranges expressed
-  with ``[]`` will be correctly matched.  Recursive includes from
-  included files are not supported.
+  with ``[]`` will be correctly matched.  The string expression is
+  evaluated against a dictionary that includes ``host_node_name``
+  and ``here`` (the directory of the supervisord config file).  Recursive
+  includes from included files are not supported.
 
   *Default*: No default (required)
 
   *Required*:  Yes.
 
   *Introduced*: 3.0
+
+  *Changed*: 3.3.0.  Added support for the ``host_node_name`` expansion.
 
 ``[include]`` Section Example
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


psigen:
> there are many situations where a docker container will stop but will not be removed.

Just curious, is this case somehow not handled by the docker `--rm` flag?  That's how I have been handling this situation myself and haven't seen any zombie containers.

https://docs.docker.com/engine/reference/run/#clean-up-rm

maihde:
Here are the ones I've encountered:

a) The machine is rebooted
b) The docker daemon is shutdown

Otherwise, I agree with you, the --rm flag would be preferred.

Michael

psigen:
Good point w/ those use cases.

I have been following the trick used by CoreOS `fleet` launch scripts (which use a `supervisor`-link interface) in which the pre-launch command is to remove any existing container:

```ini
[Unit]
Description=MyApp
After=docker.service
Requires=docker.service

[Service]
TimeoutStartSec=0
ExecStartPre=-/usr/bin/docker kill busybox1
ExecStartPre=-/usr/bin/docker rm busybox1
ExecStartPre=/usr/bin/docker pull busybox
ExecStart=/usr/bin/docker run --name busybox1 busybox /bin/sh -c "trap 'exit 0' INT TERM; while true; do echo Hello World; sleep 1; done"
ExecStop=/usr/bin/docker stop busybox1
```
https://coreos.com/fleet/docs/latest/launching-containers-fleet.html

maihde:
Yeah, that would work as well (i.e. having supervisor provide a pre-launch
command).

~Michael

vladfr:
@maihde @psigen what would be the disadvantages of just using the existing "command" arg?
Do you see any benefits of implementing a pre-command?

Should look something like this:
```
docker ps --all --format {{.Names}} | grep ^myContainer$ \
  && docker stop myContainer && docker rm myContainer
docker run --rm --name myContainer my/image
```

My guess is that you'd need to trap & pass TERM, since this would be a bash script starting the `docker` process (as mentioned in #772 and http://veithen.github.io/2014/11/16/sigterm-propagation.html)

psigen:
@vladfr I don't think there is any particular advantage to the pre-launch command here other than it exists in the `fleet` lifecycle already. `fleet` [copies this lifecycle](https://coreos.com/docs/launching-containers/launching/getting-started-with-systemd/#advanced-unit-files) from `systemd`.

However, note that this lifecycle has some advantages w.r.t dependency management that might relate to your work here https://github.com/Supervisor/supervisor/issues/122 (_:+1: thank you for looking into this!_).

There are cases where a failure during `pre-launch` can efficiently _preempt_ the launch of dependent processes. I suspect this is one reason the stage exists in `systemd`: [for its `Before/After/Requires/Wants` dependency semantics](https://www.digitalocean.com/community/tutorials/understanding-systemd-units-and-unit-files#unit-section-directives) (which are definitely worth studying when dealing with #122).

Issue #749, opened 2016-04-29
Line breaks in logs are ignored because of fixed length buffer


neumayer:
Currently supervisor reads a fixed lenght buffer from stdout. The content of this buffer is then flushed to files (if set up to do so). This leads to files ending in or starting with incomplete log lines, which comes apparent when logrotation is on.

This is a problem for every setup that wants to use supervisor to redirect a process' stdout to files and these files to be shipped off by another process (i.e. a logstash setup). I want to use a file based buffer to be more in control of network outages.

I have a commit adding a new mode (similar to the capture mode) that respects line breaks before flushing via buffer manipulation (i.e. keep text after the last line break in the buffer for the next flush operation. This also means that this new mode is independent from capture mode, i.e. capture mode won't work with respecting newlines.

Any pointers, opinions, ideas? Honestly, I'm a bit surprised that no one else has experienced this before (at least according to "the Internet"). Has this ever been discussed before?

neumayer:
https://github.com/Supervisor/supervisor/pull/751

mnaberez:
Closing this in favor of #751.

PR #751, opened 2016-04-29
Add mode to honor linebreaks in stdout

neumayer:
In order to avoid line breaks in the middle of log lines (which
currently happens when the buffer is flushed) we need a variable length
buffer. This is solved by flushing only up to the next line break and
keeping the remainder of the current buffer until the next flush.

In this commit a config parameter for this is added:
honor_log_linebreaks=true

The default value is false to not break existing behaviour.

### DIFF
--- a/supervisor/dispatchers.py
+++ b/supervisor/dispatchers.py
@@ -84,6 +84,7 @@ class POutputDispatcher(PDispatcher):
     capturelog = None # the logger while we're in capturemode
     childlog = None # the current logger (event or main)
     output_buffer = '' # data waiting to be logged
+    linebreakmode = False
 
     def __init__(self, process, event_type, fd):
         """
@@ -120,6 +121,7 @@ class POutputDispatcher(PDispatcher):
         self.log_to_mainlog = config.options.loglevel <= self.mainlog_level
         self.stdout_events_enabled = config.stdout_events_enabled
         self.stderr_events_enabled = config.stderr_events_enabled
+        self.linebreakmode = config.honor_log_linebreaks
 
     def _setup_logging(self, config, channel):
         """
@@ -189,6 +191,18 @@ class POutputDispatcher(PDispatcher):
                     )
 
     def record_output(self):
+        if self.linebreakmode:
+            if "\n" in self.output_buffer:
+                split = self.output_buffer.rsplit("\n", 1)
+                data  = split[0] + "\n"
+                self.output_buffer = split[1]
+            else:
+                data = self.output_buffer
+                self.output_buffer = ''
+
+            self._log(data)
+            return
+
         if self.capturelog is None:
             # shortcut trying to find capture data
             data = self.output_buffer
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -763,6 +763,7 @@ class ServerOptions(Options):
                 continue
             program_name = process_or_group_name(section.split(':', 1)[1])
             priority = integer(get(section, 'priority', 999))
+
             fcgi_expansions = {'program_name': program_name}
 
             # find proc_uid from "user" option
@@ -888,6 +889,7 @@ class ServerOptions(Options):
         stdout_events = boolean(get(section, 'stdout_events_enabled','false'))
         stderr_cmaxbytes = byte_size(get(section,'stderr_capture_maxbytes','0'))
         stderr_events = boolean(get(section, 'stderr_events_enabled','false'))
+        honor_log_linebreaks = boolean(get(section, 'honor_log_linebreaks', 'false'))
         serverurl = get(section, 'serverurl', None)
         if serverurl and serverurl.strip().upper() == 'AUTO':
             serverurl = None
@@ -992,6 +994,7 @@ class ServerOptions(Options):
                 stderr_logfile=logfiles['stderr_logfile'],
                 stderr_capture_maxbytes = stderr_cmaxbytes,
                 stderr_events_enabled = stderr_events,
+                honor_log_linebreaks=honor_log_linebreaks,
                 stderr_logfile_backups=logfiles['stderr_logfile_backups'],
                 stderr_logfile_maxbytes=logfiles['stderr_logfile_maxbytes'],
                 stderr_syslog=logfiles['stderr_syslog'],
@@ -1809,7 +1812,7 @@ class ProcessConfig(Config):
         'stderr_logfile_backups', 'stderr_logfile_maxbytes',
         'stderr_events_enabled', 'stderr_syslog',
         'stopsignal', 'stopwaitsecs', 'stopasgroup', 'killasgroup',
-        'exitcodes', 'redirect_stderr' ]
+        'exitcodes', 'redirect_stderr', 'honor_log_linebreaks' ]
     optional_param_names = [ 'environment', 'serverurl' ]
 
     def __init__(self, options, **params):
--- a/supervisor/skel/sample.conf
+++ b/supervisor/skel/sample.conf
@@ -89,6 +89,12 @@ serverurl=unix:///tmp/supervisor.sock ; use a unix:// URL  for a unix socket
 ;environment=A="1",B="2"       ; process environment additions (def no adds)
 ;serverurl=AUTO                ; override serverurl computation (childutils)
 
+;[program:foohonorloglinebreaks]
+;stdout_logfile_maxbytes=1MB
+;stdout_logfile_backups=6
+;honor_log_linebreaks=true
+;command=foo
+
 ; The below sample eventlistener section shows all possible
 ; eventlistener subsection values, create one or more 'real'
 ; eventlistener: sections to be able to handle event notifications
--- a/supervisor/tests/base.py
+++ b/supervisor/tests/base.py
@@ -514,6 +514,7 @@ class DummyPConfig:
                  stdout_syslog=False,
                  stderr_logfile=None, stderr_capture_maxbytes=0,
                  stderr_events_enabled=False,
+                 honor_log_linebreaks=False,
                  stderr_logfile_backups=0, stderr_logfile_maxbytes=0,
                  stderr_syslog=False,
                  redirect_stderr=False,
@@ -531,6 +532,7 @@ class DummyPConfig:
         self.stdout_logfile = stdout_logfile
         self.stdout_capture_maxbytes = stdout_capture_maxbytes
         self.stdout_events_enabled = stdout_events_enabled
+        self.honor_log_linebreaks = honor_log_linebreaks
         self.stdout_logfile_backups = stdout_logfile_backups
         self.stdout_logfile_maxbytes = stdout_logfile_maxbytes
         self.stdout_syslog = stdout_syslog
--- a/supervisor/tests/test_dispatchers.py
+++ b/supervisor/tests/test_dispatchers.py
@@ -182,6 +182,28 @@ class POutputDispatcherTests(unittest.TestCase):
              "'process1' stdout output:\na")
         self.assertEqual(dispatcher.output_buffer, '')
 
+    def test_record_output_log_linebreakmode(self):
+        # stdout/stderr goes to the process log and the main log,
+        # in non-capturemode, the data length doesn't matter
+        options = DummyOptions()
+        from supervisor import loggers
+        options.loglevel = loggers.LevelsByName.TRAC
+        config = DummyPConfig(options, 'process1', '/bin/process1',
+                              stdout_logfile='/tmp/foo', honor_log_linebreaks=True)
+        process = DummyProcess(config)
+        dispatcher = self._makeOne(process)
+        dispatcher.output_buffer = 'a\nb'
+        dispatcher.record_output()
+        self.assertEqual(dispatcher.childlog.data, ['a\n'])
+        self.assertEqual(options.logger.data[0],
+             "'process1' stdout output:\na\n")
+        self.assertEqual(dispatcher.output_buffer, 'b')
+        dispatcher.record_output()
+        self.assertEqual(dispatcher.childlog.data, ['a\n', 'b'])
+        self.assertEqual(dispatcher.output_buffer, '')
+        self.assertEqual(options.logger.data[1],
+             "'process1' stdout output:\nb")
+
     def test_record_output_emits_stdout_event_when_enabled(self):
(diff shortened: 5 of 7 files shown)


mnaberez:
Rationale for this is explained in #749.

neumayer:
I'm open to suggestions here.

Still wondering why no one else has experienced this, maybe just no one tries to ship off the supervisor log files containing standard output content.

mnaberez:
> Still wondering why no one else has experienced this, maybe just no one tries to ship off the supervisor log files containing standard output content.

Other users are shipping logs but they aren't scraping the log files like [you described](https://github.com/Supervisor/supervisor/issues/749#issue-151838568).  The usual approach is to use the [eventlistener](http://supervisord.org/events.html#event-listeners-and-event-notifications) system.  You can write a small subprocess that receives notifications when a process writes to its stdout.  Your subprocess can implement whatever buffering or retrying mechanisms you like.  There are eventlistener plugins for remote syslog, logstash, and more on the [plugins page](http://supervisord.org/plugins.html).

neumayer:
Well, the whole point of my commit is:

* to have a clear separation of concerns: the process management handles log flushing and applications just log to stdout (instead of creating specific events)
* be able to use a separate process for log shipping using file buffering because all other approaches drop logs on network outages (filebeat is what I'm planning to use)

The way I understand it, the application would need to emit events and the whole point is to make the application independent, i.e. "write to stdout" should be the only requirement to applications. I don't think having using the event system should be needed here, it adds moving parts that are not needed (need for a separate process plus mangement thereof and clearly violates the separation of concerns because it requires the application to behave in a certain way). I'd like to point you to one of my favourite manifestos: http://12factor.net/logs suggests logs to be handled as an event stream, yes, but, quote "one event per line." I'm sceptical to the tight integration of events into both the application and execution environment because it creates somewhat blurry borders. Maybe I misunderstood though.

The advantage of having whatever runs the application process buffer logs in files is that one can reuse its log rotation system and as such adjust the buffer size before stuff gets shipped off. Not doing that properly is something all logging approaches have in common that don't write to files as an intermediary step (and I've investigated a few). So that plugin for example needs to have an internal buffer to be filled up in case the other end (logstash properly) is down or the network is gone. It may be handled in various ways and some do it better than others but at best it's a memory limitation, at works events get dropped from the second the network is down. Now this guarantee might not be as important to everyone as it currently is to me or to us, but I think it's a reasonable best effort to not use logs. Also I haven't specifically looked at the code of the plugin you linked, but I'd really rather use the official supported tool that sits on the defined interface (files) rather than a custom solution.

Another point is that from what I gather supervisor is often used in combination with docker. And a reasonable way of using it there is to have docker handle the piping of stdouts from containers to files. So we have to frameworks doing the same thing in a way. And I haven't checked how docker handles linebreaks in stdout but I think it should be handled as close to the actual application as possible which would be in supervisor. There we propagate the problem (I still think the buffer across linebreaks is unexpected when it comes to logs, all log processing systems, that I've ever encountered, are line based, i.e. have a line equals event assumption). The responsibility of handling logs would shift from supervisor to docker and the line break issue would be propagated and I see no reason for that. There's more info about the advantages in the docker world here: https://medium.com/@kelseyhightower/12-fractured-apps-1080c73d481c#.jf1ykeaaj

The change I'm proposing is rather small (i.e. removes half a line to some of the buffers, meaning some buffers will be slightly smaller than before). This has no negative consequences as far as I can see and I even think it would be sensible default behaviour.

Anyway, in my mind supervisor can very well perform the task of a central component in the 12 factor app world (i.e. "the execution environment"), which really is just a way of defining how to pass data in and out of an app (i.e. parameters in and logs as one line events back via stdout). The contract that's proposed there is clear and easy and I see clear advantages because of that.

mnaberez:
> The way I understand it, the application would need to emit events and the whole point is to make the application independent, i.e. "write to stdout" should be the only requirement to applications. 

Please see the [`PROCESS_LOG_STDOUT`](http://supervisord.org/events.html#process-log-stdout-event-type) event.  `supervisord` can emit a `PROCESS_LOG_STDOUT` event to its eventlisteners whenever a subprocess writes to its `stdout`.  The eventlistener subprocess can then do whatever buffering or log shipping is desired.  This is how the existing logging [plugins](http://supervisord.org/plugins.html) work.

neumayer:
I'll take another look, didn't have much time to experiment yet. I see though that the combination of PROCESS_LOG_STDOUT and a plugin can couple the linebreaks to the linebreaks in the application log.

I still think that the "one log event per line" expectation stands strong and indeed seems to be the standard case and that plugins/event listeners are a workaround for this (needing a plugin to be installed and configured). The more I think about the more I think there's no reason whatsoever to allow the linebreaks to come out of sync (the fixed buffer has no advantage I can see since it's only "almost-fixed" anyway).

neumayer:
I found time to take another look, and I might even have managed to understand your suggestion, sorry I didn't read the docs properly :-)

So what you're suggesting is to use a separate eventlistener process (which basically writes whatever comes to stdin to stderr). The logs of this eventlistener process will end up in another set of output files and the linebreaks will be synced to "when the process" writes.

I see the following disadvantages:

* an unnecessary second process plus code plus plugin installation
* a rather unintuitive workaround in disabling the logging for the actual [program] and instead rely on the [eventlistener] files
* excessive config for multiple processes (i.e. one program plus one eventlistener per process I want to run), plus keeping config in sync
* mixing of stdout and stderr (the eventlistener can only write data to stderr if I understand correctly since stdout is reserved for control characters), so what I'd be doing is to capture stdout and stdin and write it to stderr (which then is written to files by supervisor the same way a program's output would)

And this only works in case the actual process flushes on every log event (which, admittedly is reasonable). But if violated the line break scanning would have to be included in the eventlistener.

mnaberez:
> So what you're suggesting is to use a separate eventlistener process (which basically writes whatever comes to stdin to stderr).

You're currently writing files and then having some external program scrape those files and send the data somewhere.   Your comment above seems to be that you're thinking about the eventlistener the same way (an eventlistener writes to files, your external program scrapes the files).  

I'm suggesting that the way other people seem to have solved this is to have an eventlistener that does all the work (no external program scraping files).  You can have an eventlistener receive log data, buffer or process it how you like, and ship it to the endpoint.  There are example eventlisteners for remote syslog and logstash.  

> I see the following disadvantages:

You point out that an eventlistener requires another process and is configuration heavy, but your existing setup already has another process.  Presumably that process needs configuration as well.

neumayer:
The writing to files is a reasonable intermediate step if one doesn't want to drop log messages (it will be buffered in files and the reading continues from there once the network is back). I've seen no setup so far that gives similar guarantees sending directly (with existing tools, I think writing a log shipping plugin that does proper buffering and handling of network issues is a larger project I don't feel I have the resources for).

I agree that the scraping of files needs setup too, but it is outside of supervisor, i.e. the execution environment and the separation of concerns is well worth it, well, in my opinion, that is :-). So it'd be kind of the point to use the existing stdout to files functionality (which I still see as a core "execution environment" task, docker does a similar thing, I haven't checked how linebreaks are handled there, will try to report back in case I ever do). If using supervisor inside docker (as some people seem to suggest), the std would be piped twice and I think the linebreaks should be fixed closest to the actual application. I do see though, that this is an ideological debate.

I don't think my setup is even the point here, why would you ever want to do log rotation that splits up lines across files? Even if no log shipping ever happens and you look through the log rotated files on disc (manually), the split up lines are a bit of a nuisance... I honestly found that very unexpected because of the aforementioned "one line = one log event" principle, which seems to be rather common.

But I do see that you might not be into such change too much :D

mnaberez:
> I do see though, that this is an ideological debate.

Is that really a fair accusation to make considering that you weren't even aware of the eventlistener system's capabilities before I pointed them out to you?  I'm not going to apologize for not immediately applying this patch because it probably breaks core features like `supervisorctl fg`.

neumayer:
I didn't mean that in a negative way at all, sorry if it sounded like it. I do value your thoughts and input. And I already apologised for not properly reading the docs about the eventlistener system (and tried to follow up on that by properly reading it :-)).

mnaberez:
> I don't think my setup is even the point here, why would you ever want to do log rotation that splits up lines across files?

Another point along those lines is that it splits on bytes not characters, which means the existing implementation can probably split the log in the middle of a multibyte character.  That case is definitely undesirable.

mnaberez:
The main disadvantage of line buffering are that it breaks the "real time" features, e.g. I have some processes running under `supervisord` that log progress like `\r96%` `\r97%` `\r98%` `\r99%` `\r100%\n` and I can `supervisorctl tail -f` or `supervisorctl fg` those and see it update.  With line buffering, you wouldn't see any of that until the `\n`.  That is probably a minority use case compared to most logs, but `supervisord` does have these features, and people do use them.  If we did apply this, I would probably not make it the default, considering this.

neumayer:
I see, hadn't thought about that either. It currently works for bits and pieces larger than the current buffer, i.e. depends on the buffer size, and I think the read buffer size is 128k, so maybe there's not so much "real time" stuff going on there anyway? But maybe I misunderstood that now.

The practical open issues in the context of the commit I currently think about would be:

* how to support different kinds of linebreaks (not only unix ones)
* buffer overflow (i.e. what if there's no linebreaks ever or very, very long lines), probably solveable by a maxbuffer size of some kind

mnaberez:
> I see, hadn't thought about that either. It currently works for bits and pieces larger than the current buffer, i.e. depends on the buffer size, and I think the read buffer size is 128k, so maybe there's not so much "real time" stuff going on there anyway?

If you try the program [above](https://github.com/Supervisor/supervisor/pull/751#issuecomment-220696468), you will see the individual dots.  It won't be "real time" if the program buffers its own output, but once it flushes, `supervisord` should log it quickly.

> how to support different kinds of linebreaks (not only unix ones)

`supervisord` is a Unix program and makes many Unix assumptions already.  Assuming `\n` is probably fine.  It is pretty unlikely that `supervisord` will run on a non-Unix platform.  The issue could be revisited if it was ever ported.

> buffer overflow (i.e. what if there's no linebreaks ever or very, very long lines), probably solveable by a maxbuffer size of some kind

Yes, that would be needed.   A process could spew output with no newlines and that could cause memory bloat.  Nothing a process outputs should be able to crash `supervisord`. 

The last thing to look at when considering this patch would be what is the performance impact on searching/splitting newlines on every log write.

neumayer:
I checked, the buffer overflow is handled when reading the buffer in the options.py file. There we read a buffer of 128k (well, of max 128k, that is, that's why the progress dots are shown even if the buffer is not full).
My change currently only checks if a linebreak is in the buffer read from there and saves the part after the linebreak for the next round. If no linebreak is found everything until the end of the buffer is handled. This effectively limits the size to the max size that is read from the process.

I'll take a look at performance in a bit. The performance of my change should be inline with the capture mode. But I'm also interested in the comparison to the "normal" mode. I expect the python regular expression/string handling to be rather performant though (and I think this is what this comes down to).

neumayer:
Do you have any tips on how to measure performance? I've been thinking a bit and I'm not sure where to start. Have you done any other kind of performance measurement? For the eventsystem maybe?

I can come up with something generic for line splitting, but it seems like this would just be testing python's ability to split strings and that doesn't seem all too helpful...

Anyway, I am willing to use some time on int later this week, just not sure where to start.

IstvanZakar:
I wanted to propose a similar requirement.

We have the following scenario. Our processes are started by supervisord inside docker containers. The log level of supervisord set to DEBUG so the stdout of the processes ends up in the stdout of supervisord and finally in the docker logs, so we can gather it with logstash. But instead of having the current printout:

2016-06-15 10:38:43,881 DEBG 'processA' stdout output:
log line1
log line2
log line3

I would like to propose an option to have this format:
2016-06-15 10:38:43,881 DEBG 'processA' stdout output: log line1
2016-06-15 10:38:43,881 DEBG 'processA' stdout output: log line2
2016-06-15 10:38:43,881 DEBG 'processA' stdout output: log line3

Would you consider accepting an option for a change like the one I mentioned?

mnaberez:
> I would like to propose an option to have this format:
2016-06-15 10:38:43,881 DEBG 'processA' stdout output: log line1
2016-06-15 10:38:43,881 DEBG 'processA' stdout output: log line2
2016-06-15 10:38:43,881 DEBG 'processA' stdout output: log line3
>
> Would you consider accepting an option for a change like the one I mentioned?

No.  Debug log messages are intended for developers troubleshooting Supervisor internals.  We won't add special options to influence the format of debug log messages.  Debug log messages may change without warning or may be removed.

IstvanZakar:
you mean the feature of forwarding stdout of a process to the stdout of supervisord may be removed?

neumayer:
I'm not sure I see how this is connected to the linebreaks.

IstvanZakar:
If I have a log which consists of multiple lines (an exception maybe) supervisord prints one line like this:
2016-06-15 10:38:43,881 DEBG 'processA' stdout output: 
An than the whole exception (depending on the length). But if I need to process this output afterwards I cannot do that line-by-line. The suggestion would be instead of printing one "header" line have an option to print the procees information at the begining of every line.

mnaberez:
> you mean the feature of forwarding stdout of a process to the stdout of supervisord may be removed?

I don't think this is documented in the manual and the messages logged at the `DEBUG` level were not originally intended to use this way.  There's no immediate plans to remove this particular message but I don't think the project can make commitments like keeping `TRACE` or `DEBUG` messages forever, making them a specific format, etc.  They're intended for development and troubleshooting, and they will change.

You can run a plugin like [`supervisor-stdout`](https://github.com/coderanger/supervisor-stdout) that uses the eventlistener system (a published and stable interface) to send child logs to stdout.  This also avoids all the other `DEBUG` messages that you probably don't care about.  You can also modify `supervisor-stdout` or write your own to format the messages however you want without modifying `supervisord`.

> I'm not sure I see how this is connected to the linebreaks.

I don't see the connection either.  Let's stop this side discussion and leave this issue for line buffering.

IstvanZakar:
Okay, thanks for your comments.

 I brought this topic up, as my suggested modification would have been exactly the same as the one proposed in this pull request ( with one additional small change about removing a linebreak ) so in this sense I cannot see why it's not related.

mnaberez:
Closing due to inactivity.

PR #754, opened 2016-05-12
Add a program option to update oom_score_adj for child processes.

yossigo:
Configuring the oom_score_adj value for child processes is a useful way of controlling the kernel's out-of-memory killer behavior.  Other modern process managers (e.g. systemd, Upstart) also support it and it makes a lot of sense for Supervisor to do as well.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -891,6 +891,7 @@ class ServerOptions(Options):
         serverurl = get(section, 'serverurl', None)
         if serverurl and serverurl.strip().upper() == 'AUTO':
             serverurl = None
+        oom_score_adj = get(section, 'oom_score_adj', None)
 
         # find uid from "user" option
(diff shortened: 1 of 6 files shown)


mnaberez:
Added [rlimits](https://github.com/Supervisor/supervisor/labels/rlimits) issue label (it's not an rlimit but that's the closest label we have for this).  If such a feature was merged, it would be the first time a Linux-only feature was added to Supervisor.

ekbfh:
Hello! 
What else is needed to merge this code?

EscapeLife:
Hello! When will it merge? @yossigo @mnaberez

ekbfh:
@mnaberez how i can help to merge this?

guyco-redis:
@mnaberez can we promote this?

PR #760, opened 2016-05-26
Fixed race condition per https://github.com/scottp-dpaw/supervisor/pu…

mattmarich:
…ll/1/files

Fixed race condition per: https://github.com/scottp-dpaw/supervisor/pull/1/files

We've seen supervisor crash multiple times due to the 'UNKNOWN' process state, the processes continue running but they become detached from supervisor and need to be killed manually prior to supervisor restart. 

We've had this patch running live for a few weeks and we haven't seen any issues on our patched systems but continue to see the problem on our unpatched systems.

### DIFF
--- a/supervisor/process.py
+++ b/supervisor/process.py
@@ -477,7 +477,7 @@ class Subprocess(object):
             msg = 'unknown problem sending sig %s (%s):%s' % (
                                 self.config.name, self.pid, tb)
             options.logger.critical(msg)
-            self.change_state(ProcessStates.UNKNOWN)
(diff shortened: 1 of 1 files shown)


mnaberez:
Duplicate of #688.

mattmarich:
This may be a duplicate but you have yet to accept one of the pull requests. This bug makes supervisor highly unstable so it needs to be patched. The only other method would be for those experiencing the issue to maintain their own fork of supervisor so we'd miss out on any other patches. I don't care which pull request is merged but someone needs to implement a sanity check for the 'UNKNOWN' exception.

PR #776, opened 2016-06-24
add 'dependson' option for program and group

c0b:
Dependson implements dependency management for [program] and [group]
a program or group won't start until all its dependencies are in RUNNING state.
Note that the group-level setting overrides the program-level setting:
- as it does not make sense for 2 programs in the same group to depend on different things;
- and in supervisor, a program is a ProcessGroup

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -673,12 +673,14 @@ class ServerOptions(Options):
             return parser.saneget(section, opt, default, **kwargs)
 
         # process heterogeneous groups
+        program_and_group_names = []
         for section in all_sections:
             if not section.startswith('group:'):
                 continue
             group_name = process_or_group_name(section.split(':', 1)[1])
             programs = list_of_strings(get(section, 'programs', None))
             priority = integer(get(section, 'priority', 999))
+            dependson = list_of_strings(get(section, 'dependson', None))
             group_processes = []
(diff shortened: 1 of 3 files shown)


mnaberez:
[Opened by mistake](https://github.com/Supervisor/supervisor/issues/122#issuecomment-228428559).

rrei:
Is there any logic reason why this is not being merged?! It's been 4 years and dozens of people requesting this feature... I'm one of them. This is frustrating.

mnaberez:
> Is there any logic reason why this is not being merged?!

A person who was not the author of this patch submitted it and then said it was [opened by mistake](https://github.com/Supervisor/supervisor/issues/122#issuecomment-228416456), which is why I closed it.  The author of the patch hasn't submitted it and [also said it isn't done](https://github.com/Supervisor/supervisor/issues/122#issuecomment-228416456).

rrei:
Alright, that makes sense, but still... why doesn't any of the organization members (such as @mnaberez) pick up this seemingly not so complicated task and just be done with it? A looooot of people have been asking for this for a loooooooong time. I see that new releases are being launched fairly regularly, so **someone** must be working on the code, right?
From a developer's perspective, this sounds like it shouldn't be too difficult to implement. Am I wrong?

PR #784, opened 2016-07-12
Exit if args given to reload

jkrauska:
Users can accidentally confuse reload (reload the entire supervisor and all children) with restart (restart a single child).  This keeps reload from working if a trailing argument is given.

### DIFF
--- a/supervisor/supervisorctl.py
+++ b/supervisor/supervisorctl.py
@@ -983,6 +983,11 @@ class DefaultControllerPlugin(ControllerPluginBase):
         self.ctl.output("shutdown \tShut the remote supervisord down.")
 
     def do_reload(self, arg):
+        if arg:
+            self.ctl.output('Error: reload given with a process name. Maybe you meant restart?')
+            self.help_reload()
+            return
+ 
         if self.ctl.options.interactive:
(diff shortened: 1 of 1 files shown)


supervisor/supervisorctl.py
```diff
@@ -983,6 +983,11 @@ def help_shutdown(self):
         self.ctl.output("shutdown \tShut the remote supervisord down.")
 
     def do_reload(self, arg):
+        if arg:
+            self.ctl.output('Error: reload given with a process name. Maybe you meant restart?')
```
mnaberez: This should use `handle_error` [like the other](https://github.com/Supervisor/supervisor/blob/f4c55bacd502d46652444cc20405121a75d0762b/supervisor/supervisorctl.py#L831-L832) argument checks in this file.

mnaberez:
Thanks for this patch, I will try to include this in the next point release.  It will need a unit test before merging (most or all of the existing arg checks have one).  Feel free to add a test or I'll take care of it later.

mnaberez:
Committed in 51caeb7eed819dceacf86e988739e022c7f78cd1.

PR #789, opened 2016-07-18
Fix #787

### LINKED ISSUE #787: Searching for command in PATH ignores changes made by environment=
I write a supervisord.conf，program section like this:

```
[program:xxx]
environment=PATH=%(ENV_HOME)s/.virtualenvs/venv/bin:%(ENV_PATH)s
command=gunicorn -w 3 xxx.app:app -b 0.0.0.0:8000
...
```

but it's fail. supervisor told me:

```
can't find command 'gunicorn'
```

but i can find it:

```
❯ ll ~/.virtualenvs/venv/bin/gunicorn 
-rwxrwxr-x 1 ubuntu ubuntu 245 Jun 12 05:39 /home/ubuntu/.virtualenvs/venv/bin/gunicorn
```

Does I used the wrong way?

dongweiming:
ref: https://github.com/Supervisor/supervisor/issues/787#issuecomment-233224771

@mnaberez. I know why can't find command 'gunicorn'

This error raise at https://github.com/Supervisor/supervisor/blob/master/supervisor/process.py#L216

because before execute [get_path method](https://github.com/Supervisor/supervisor/blob/master/supervisor/process.py#L126),  env is not updated, so `get_path()`"s result not contains `%(ENV_HOME)s/.virtualenvs/venv/bin`

You can create a simple example with virtualenv env to find out this.

### DIFF
--- a/supervisor/process.py
+++ b/supervisor/process.py
@@ -123,6 +123,8 @@ class Subprocess(object):
                 st = None
 
         else:
+            if self.config.environment is not None:
+                os.environ.update(self.config.environment)
             path = self.config.options.get_path()
             found = None
             st = None
@@ -923,6 +925,3 @@ def new_serial(inst):
         inst.serial = -1
     inst.serial += 1
     return inst.serial
-
-
-


supervisor/process.py
```diff
@@ -123,6 +123,8 @@ def get_execv_args(self):
                 st = None
 
         else:
+            if self.config.environment is not None:
+                os.environ.update(self.config.environment)
```
mnaberez: The function you are changing, `get_execv_args()`, [runs in the `supervisord` process](https://github.com/Supervisor/supervisor/blob/154cb4c84f28ac3e0ac1ce5409faea65c15d2d02/supervisor/rpcinterface.py#L279).  If this change was made, the `environment=` of any `[program:x]` section would leak into `supervisord` and then into all subprocesses.  The expected behavior is that the `environment=` of a `[program:x]` affects only that program.

dongweiming:
@mnaberez I'm not sure a better way, but this bug is exists. could you fix it?

mnaberez:
Reopened #787.

PR #812, opened 2016-08-19
Fixed logfile output to special files like /dev/stdout on Python 3.

hathawsh:
Python 3.5.2 disallows `open('/dev/stdout', 'a')` (it causes OSError with errno=ESPIPE) but allows `open('/dev/stdout', 'w')`. This change makes Supervisor detect this error condition and fix it if possible.

(I might try to get this fixed in Python itself, but this patch helps Supervisor work with existing Python 3 installations.)

### DIFF
--- a/supervisor/loggers.py
+++ b/supervisor/loggers.py
@@ -9,7 +9,6 @@ idiosyncratic and a bit slow for our purposes (we don't use threads).
 
 import os
 import errno
-import stat
 import sys
 import time
 import traceback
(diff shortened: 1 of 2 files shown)


supervisor/loggers.py
```diff
@@ -138,11 +139,29 @@ class FileHandler(Handler):
...
+    def __init__(self, filename, mode=None):
         Handler.__init__(self)
-        self.stream = open(filename, mode)
+        open_mode = mode or 'a'
+
+        try:
+            self.stream = open(filename, open_mode)
+        except OSError as e:
+            retried = False
+            if not mode:
```
mnaberez: I took me a couple passes through but I think what this dual `mode` and `open_mode` is intended to do is "If nothing was given for mode, try 'a' and then try 'w' if the Python 3 bug occurs.  But if a specific mode was supplied, try only that mode and don't work around the Python 3 bug".

Is there a case where having the workaround for the Python 3 implicit SEEK would be undesirable?  If `'w'` and `'a'` really do have the same semantics on these devices, we can simplify this back to `mode` and remove the special `if not mode` case.  I find the dual `mode`/`open_mode` pretty confusing.  If we do need to keep it, we should probably add some comments explaining why we need the separate behaviors.

hathawsh: Yes, I wrote it that way out of an abundance of caution. However, I'm
learning from the parallel Python bug that the C library code used by
Python 2 is actually rather naive: it merely swallows any seek error when
opening in append mode. We can emulate that behavior with simpler code,
like you suggest. I'll update the pull request.

hathawsh:
The pull request is now simplified. Thanks for the feedback.

mnaberez:
Thanks!

PR #816, opened 2016-08-29
Correct redirect_stderr target in 3.0a1 release note

tedder:
Changing history is frowned upon, but this change doesn't match what the code actually does (nor does it match the documentation). Since it could be confusing, I think it's definitely worth fixing.


tedder:
Failing from an unrelated error on one of the test envs.

PR #842, opened 2016-10-28
Move from easy_install to pip. Fixes #623.

### LINKED ISSUE #623: docs mention installation with `easy_install`; should be replaced with `pip`
http://supervisord.org/installing.html

the page is confusing, since it says "if you've got internet access, install using `easy_install`"

Page should be updated to use `pip` not `easy_install`, or at the very least, demote `easy_install` to be below `pip`

jbenjos:


mnaberez:
Closing due to inactivity.

Issue #827, opened 2016-09-20
Supervisord web interface behind proxy with HTTPS scheme


ulrith:
It would be great if we able correctly use Supervisord web interface behind proxy with HTTPS scheme. Currently Supervisord controls return Location header with HTTP scheme (which can be easily redirected to HTTPS). But if it resides inside iframe it's blocked by modern browsers (mixed content error).

Apparent solution seems to give more power to user for control of Location header through configuration settings (url parameter?) or use "X-Forwarded-Proto" or "X-Forwarded-Protocol" header as @fengsi [suggested](https://github.com/Supervisor/supervisor/issues/251#issuecomment-32250649).

mnaberez:
Duplicate of #29, #251.

PR #845, opened 2016-11-04
Fix reverse proxy message redirect after web action

razvanphp:
This small change allows supervisor HTTP server to be served behind a reverse proxy like nginx by fixing the redirect (the only one in web.py).

This is temporary until version 4.0 is launched with the configurable `base_path` (#593), given the release cycles are so big & there are so much left to do for the python 3 release.

This works very well behind nginx HTTPS offloading without any hack in the configuration:
```
        location /supervisor/job/ {
                proxy_pass http://10.107.0.10:9001/;
                include includes/proxy-parameters;
        }
```

I know you already refused this once in PR #495, but I would kindly ask you to reconsider this, it's not breaking anything, there are already only relative links in the whole web interface, so I can't see why this one should not be. This way you actually don't need any `base_path` anymore.

Related to #29, #827, makes #669 redundant.

### DIFF
--- a/supervisor/web.py
+++ b/supervisor/web.py
@@ -462,8 +462,7 @@ class StatusView(MeldView):
                 if message is NOT_DONE_YET:
                     return NOT_DONE_YET
                 if message is not None:
-                    server_url = form['SERVER_URL']
-                    location = server_url + '?message=%s' % urllib.quote(
+                    location = 'index.html?message=%s' % urllib.quote(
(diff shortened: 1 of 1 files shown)


mnaberez:
> I know you already refused this once in PR #495, but I would kindly ask you to reconsider this

Please don't open identical duplicate PRs.  The place to discuss #495 is on #495.

PR #849, opened 2016-11-14
Support forwarding stdout of processes to supervisor stdout

wmanley:
This supports a common container use-case where supervisor is pid 1 in a
container and the container manager will be collecting and recording the
output of supervisor.  Users can now specify:

    stdout_logfile=stdout
    stderr_logfile=stderr

and supervisor will pass any messages logged strieght through.

### DIFF
--- a/supervisor/loggers.py
+++ b/supervisor/loggers.py
@@ -395,7 +395,10 @@ def handle_syslog(logger, fmt):
 def handle_file(logger, filename, fmt, rotating=False, maxbytes=0, backups=0):
     if filename == 'syslog':
         handler = SyslogHandler()
-
+    elif filename == 'stdout':
+        handler = StreamHandler(sys.stdout)
+    elif filename == 'stderr':
+        handler = StreamHandler(sys.stderr)
     else:
         if rotating is False:
             handler = FileHandler(filename)


mnaberez:
It's pretty easy to achieve this without modifying `supervisord`, and the usage is almost the same:
http://veithen.github.io/2015/01/08/supervisord-redirecting-stdout.html

There's also a plugin:
https://github.com/coderanger/supervisor-stdout

wmanley:
> It's pretty easy to achieve this without modifying supervisord, and the usage is almost the same:
http://veithen.github.io/2015/01/08/supervisord-redirecting-stdout.html

Previously I had been using this workaround, but it doesn't work when you don't have permission to open `/dev/stdout`.  This is the case with containers started with `systemd-nspawn`.  This is why I created this patch.

mnaberez:
> Previously I had been using this workaround, but it doesn't work when you don't have permission to open /dev/stdout. This is the case with containers started with systemd-nspawn. This is why I created this patch.

The plugin linked above won't have that problem.

flavius:
@mnaberez The plugin doesn't seem to work any more.

mnaberez:
> The plugin doesn't seem to work any more.

Please report issues with the plugin on [its issue tracker](https://github.com/coderanger/supervisor-stdout).

PR #854, opened 2016-11-23
Enable sending signals to stopping processes

mikeyg123:
Add a set of SIGNALABLE_STATES and use this to determine if a signal can be sent to a process.


mikeyg123:
it's not up to me unfortunately
As I didn't want to maintain my own fork I got around this issue another way.  The supercast plugin we built includes the ability to monkey-patch (replace at runtime) a function in the rpc interface. 
See the last few lines from about 584 here https://git.tudor.com/mgould/supercast/blob/master/supercast.py

Issue #863, opened 2016-12-12
[supervisord] and [supervisorctl] config sections are not optional


lukassup:
Reading the [documentation for 3.3.1](http://supervisord.org/running.html#adding-a-program) it says that it should be enough to add:
```ini
[program:foo]
command=/bin/cat
```
to the `supervisord.conf` file to run `supervisord`.

However with the master branch I am not able to do that. It says that `[supervisord]` section is required in the `supervisord.conf`. Running `supervisorctl` says `[supervisorctl]` section is required as well.

Since every option has defaults already why not make the whole section optional as well?

mnaberez:
> However with the master branch I am not able to do that. 

Every version of `supervisord` has required the `[supervisord]` section in the config file to my knowledge, so it's not a regression on the master branch.

lukassup:
#864 pull request to make this happen

mnaberez:
I don't think we should make these sections optional.  If nothing else, having these checks at least guarantees that you have pointed `supervisord` at the correct config file.  There's no benefit to making them optional, other than allowing a single line in the config file to be deleted.  I'd prefer to leave this alone, sorry.

lukassup:
Thanks for the heads up @mnaberez. This is definitely not a big of a deal just a 'nice-to-have' feature.

PR #864, opened 2016-12-12
Make [supervisord] and [supervisorctl] config sections optional

lukassup:


mnaberez:
Response in https://github.com/Supervisor/supervisor/issues/863#issuecomment-266337448.

Issue #470, opened 2014-07-28
docs don't list the supervisorctl commands


jvanasco:
The docs, and the website, don't have any info on the supervisord commands:
- http://supervisord.org/index.html
  
  supervisor> help
  # default commands (type help <topic>):
  
  add    clear  fg        open  quit    remove  restart   start   stop  update 
  avail  exit   maintail  pid   reload  reread  shutdown  status  tail  version

i suggest putting the (docstrings?) text from the various `help <topic>` into the "Running supervisorctl" section [http://supervisord.org/running.html#running-supervisorctl]

PR #866, opened 2016-12-22
Document supervisorctl actions

jpcaruana:
relates to #250 and #470

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -166,6 +166,27 @@ value in the configuration file.
    the supervisord process before it will start successfully.
 
 
+Running :program:`supervisorctl`
+--------------------------------
+
+To start :program:`supervisorctl`, run ``$BINDIR/supervisorctl``.  A
+shell will be presented that will allow you to control the processes
+that are currently managed by :program:`supervisord`.  Type "help" at
+the prompt to get information about the supported commands.
+
+The :command:`supervisorctl` executable may be invoked with "one time"
+commands when invoked with arguments from a command line.  An example:
+``supervisorctl stop all``.  If arguments are present on the
+command-line, it will prevent the interactive shell from being
+invoked.  Instead, the command will be executed and ``supervisorctl``
+will exit with a code of 0 for success or running and non-zero for
+error. An example: ``supervisorctl status all`` would return non-zero
+if any single process was not running.
+
+If :command:`supervisorctl` is invoked in interactive mode against a
+:program:`supervisord` that requires authentication, you will be asked
+for authentication credentials.
+
 :command:`supervisorctl` Command-Line Options
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


docs/running.rst
```diff
@@ -205,26 +226,134 @@ interactively is started.  Use the action "help" to find out about available
...
+
+   Print help for <action>
+
+add <name> [...]
+
+   Activates any updates in config for process/group
+
+remove <name> [...]
+
+   Removes process/group from active confi
```
VynceMontgomery: presume you mean "config"

docs/running.rst
```diff
@@ -205,26 +226,134 @@ interactively is started.  Use the action "help" to find out about available
...
+
+update
+
+   Reload config and add/remove as necessary
+
+clear <name>
+
+   Clear a process' log files.
```
VynceMontgomery: Prefer: Clear the log files associated with process named <name>.

docs/running.rst
```diff
@@ -205,26 +226,134 @@ interactively is started.  Use the action "help" to find out about available
...
+
+clear <name> <name>
+
+   Clear multiple process' log files
```
VynceMontgomery: Combine with above. Add "Multiple names can be provided; the log files for each process will be removed."

docs/running.rst
```diff
@@ -205,26 +226,134 @@ interactively is started.  Use the action "help" to find out about available
...
+
+clear all
+
+   Clear all process' log files
```
VynceMontgomery: Would prefer something more explicit, like "Clear log files for all processes managed by supervisor".

docs/running.rst
```diff
@@ -205,26 +226,134 @@ interactively is started.  Use the action "help" to find out about available
...
+   Note: restart does not reread config files. For that, see reread and update.
+
+restart all
+
+   Restart all processes
+   Note: restart does not reread config files. For that, see reread and update.
+
+signal
+
+   No help on signal
```
VynceMontgomery: Really? None at all? Does it do nothing?

docs/running.rst
```diff
@@ -205,26 +226,134 @@ interactively is started.  Use the action "help" to find out about available
...
+
+start <name>
+
+   Start a process
```
VynceMontgomery: what is the effective difference between start and restart? Will start no-op if the process is running?  will restart fail if the process is not?

docs/running.rst
```diff
@@ -205,26 +226,134 @@ interactively is started.  Use the action "help" to find out about available
...
+
+stop all
+
+   Stop all processes
+
+tail [-f] <name> [stdout|stderr] (default stdout)
+
+   Output the last part of process logs
+   Ex:
+   tail -f <name>		Continuous tail of named process stdout Ctrl-C to exit.
```
VynceMontgomery: use of term "follow along" will help people understand mnemonic for -f. Add semicolon or period before "Ctrl-C".  May also mention that tail is similar to or inspired by the program of the same name.

docs/running.rst
```diff
@@ -205,26 +226,134 @@ interactively is started.  Use the action "help" to find out about available
...
+   tail -100 <name>	last 100 *bytes* of process stdout
```
VynceMontgomery: -<N> does not appear in command signature. Is it compatible with -f?

docs/running.rst
```diff
@@ -205,26 +226,134 @@ interactively is started.  Use the action "help" to find out about available
...
+   Get the PID of a single child process by name.
+
+pid all
+
+   Get the PID of every child process, one per line.
+
+restart <name>
+
+   Restart a process
+   Note: restart does not reread config files. For that, see reread and update.
```
VynceMontgomery: "reread" and "update" are both missing from this list. Are they the same?  what's the difference?  Do they imply a restart, or do you have to reread and then restart?

Issue #836, opened 2016-10-09
supervisorctl fg always raises INCORRECT_PARAMETERS on Python 3


mnaberez:
Start `supervisord` in the foreground with a subprocess running:

```
$ python --version
Python 3.5.2

$ cat supervisord.conf
[supervisord]
loglevel = blather
nodaemon = true

[inet_http_server]
port = 127.0.0.1:9001

[supervisorctl]
serverurl = http://127.0.0.1:9001

[rpcinterface:supervisor]
supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface

[program:cat]
command = /bin/cat

$ supervisord -c supervisord.conf
```

In another terminal, start `supervisorctl fg` and type something followed by return:

```
$ supervisorctl -c supervisord.conf fg cat
==> Press Ctrl-C to exit <==
hi
ERROR: <Fault 2: 'INCORRECT_PARAMETERS'>
Exiting foreground
$
```

Tested on 427eb2bc6b08f788573deb91d1391d93f8b58a1b.  Note that in earlier versions, `fg` swallowed most XML-RPC faults so it would fail silently.  The cause of `INCORRECT_PARAMETERS` is the same as https://github.com/Supervisor/supervisor/issues/835#issue-181898879.   Once that is fixed, we should try the above again to be sure `supervisorctl fg` is also fixed.

JockDaRock:
what is the progress of getting supervisor to play well with Python 3???

mnaberez:
> what is the progress of getting supervisor to play well with Python 3???

Issues that still need to be resolved are tagged [python 3](https://github.com/Supervisor/supervisor/labels/python%203).

AlJohri:
if I wanted to support supervisor getting python3 support via [bountysource](https://www.bountysource.com/) which issue would be a good one to donate to? @mnaberez

mnaberez:
> if I wanted to support supervisor getting python3 support via bountysource 

This project is not current participating in any bounty websites and is also not currently soliciting monetary donations.  If the project participates in any bounties or seeks donations, it will be indicated on [supervisord.org](http://supervisord.org) or the [supervisor-users](https://lists.supervisord.org/mailman/listinfo/supervisor-users) mailing list.  Please do not enroll the project in such websites yourself; the project developers can do this themselves if they decide to do so.

AlJohri:
cool, thanks for the heads up!

Issue #835, opened 2016-10-09
sendProcessStdin() always raises INCORRECT_PARAMETERS on Python 3


mnaberez:
Start `supervisord` in the foreground with a subprocess running:

```
$ python --version
Python 3.5.2

$ cat supervisord.conf
[supervisord]
loglevel = blather
nodaemon = true

[inet_http_server]
port = 127.0.0.1:9001

[supervisorctl]
serverurl = http://127.0.0.1:9001

[rpcinterface:supervisor]
supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface

[program:cat]
command = /bin/cat

$ supervisord -c supervisord.conf
```

In another terminal, try the XML-RPC method `sendProcessStdin`:

```
$ python
Python 3.5.2 (default, Sep 10 2016, 08:21:44)
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from xmlrpc.client import ServerProxy
>>> server = ServerProxy('http://127.0.0.1:9001/RPC2', verbose=True)
>>> server.supervisor.sendProcessStdin('cat', 'hello')
send: b'POST /RPC2 HTTP/1.1\r\nHost: 127.0.0.1:9001\r\nAccept-Encoding: gzip\r\nContent-Type: text/xml\r\nUser-Agent: Python-xmlrpc/3.5\r\nContent-Length: 229\r\n\r\n'
send: b"<?xml version='1.0'?>\n<methodCall>\n<methodName>supervisor.sendProcessStdin</methodName>\n<params>\n<param>\n<value><string>cat</string></value>\n</param>\n<param>\n<value><string>hello</string></value>\n</param>\n</params>\n</methodCall>\n"
reply: 'HTTP/1.1 200 OK\r\n'
header: Content-Type header: Content-Length header: Date header: Server body: b"<?xml version='1.0'?>\n<methodResponse>\n<fault>\n<value><struct>\n<member>\n<name>faultString</name>\n<value><string>INCORRECT_PARAMETERS</string></value>\n</member>\n<member>\n<name>faultCode</name>\n<value><int>2</int></value>\n</member>\n</struct></value>\n</fault>\n</methodResponse>\n"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/xmlrpc/client.py", line 1092, in __call__
    return self.__send(self.__name, args)
[...]

Issue #663, opened 2015-10-06
Event listeners crash supervisord on Python 3


palmkevin:
I get the following error when running the master with Python 3.2:

```
 File "/opt/ls/lx/env/d/python32/supervisor/process.py", line 825, in dispatch
    ok = self._dispatchEvent(event)
  File "/opt/ls/lx/env/d/python32/supervisor/process.py", line 873, in _dispatchEvent
    process.write(as_bytes(envelope))
  File "/opt/ls/lx/env/d/python32/supervisor/process.py", line 100, in write
    dispatcher.input_buffer += chars
TypeError: Can't convert 'bytes' object to str implicitly
```

When I do replace in process.py (line 873) `as_bytes` by `as_string`, then it is working...

Pehat:
Consider using [six](https://pypi.python.org/pypi/six) package, so you would have the same-named byte-string conversion functions for Python 2 and Python 3 code.

mnaberez:
> Consider using six package

Supervisor has its own [`compat.py`](https://github.com/Supervisor/supervisor/blob/c28db88f75b916ea4d91c114e2bcb654ecc081be/supervisor/compat.py) module for Python 2/3 shims.  If any more are needed, they can can be added to `compat.py`.  Adding more package dependencies to Supervisor is undesirable, and `compat.py` is already there.

mnaberez:
Instructions to reproduce [the crash above](https://github.com/Supervisor/supervisor/issues/663#issue-109944572):

```
$ python --version
Python 3.5.2

$ cat supervisord.conf
[supervisord]
loglevel=blather
nodaemon=true

[eventlistener:cat]
command=bash -c 'echo READY; exec cat'
events=TICK_5

$ supervisord -c supervisord.conf
2016-10-08 22:18:30,370 INFO Increased RLIMIT_NOFILE limit to 1024
2016-10-08 22:18:30,373 INFO supervisord started with pid 15411
2016-10-08 22:18:31,376 INFO spawned: 'cat' with pid 15414
2016-10-08 22:18:31,384 BLAT read event caused by <PEventListenerDispatcher at 4425371432 for <Subprocess at 4436932312 with name cat in state STARTING> (stdout)>
2016-10-08 22:18:31,385 DEBG 'cat' stdout output:
READY

2016-10-08 22:18:31,385 DEBG cat: ACKNOWLEDGED -> READY
[...]

Issue #638, opened 2015-07-21
unicode decoding error while reading from stderr / stdout


peick:
Having a supervisor configuration that produces non-utf-8 data (e.g. binary) triggers an error in supervisor followed by a close of stdout / stderr for that process. This leads to a not responding process because data which should be send to stdout / stderr are blocked / stuck in the write operation.

```
[program:produce-unicode-error]
command=echo -e "\x88"
```

Log message of supervisor (formatted):

```
2015-07-21 15:56:04,934 CRIT uncaptured python exception, closing channel <POutputDispatcher at 26310992 for <Subprocess at 26313168 with name produce-unicode-error in state STARTING>  
    (stdout)> (<type 'exceptions.UnicodeDecodeError'>:'utf8' codec can't decode byte 0x88 in position 0: invalid start byte
    [/opt/python2.6/supervisor/supervisord.py|runforever|223] 
    [/opt/python2.6/supervisor/dispatchers.py|handle_read_event|260] 
    [/opt/python2.6/supervisor/options.py|readfd|1507]
    [/opt/python2.6/supervisor/compat.py|as_string|24]
    [/opt/python2.6/encodings/utf_8.py|decode|16])
2015-07-21 15:56:04,934 DEBG fd 6 closed, stopped monitoring <POutputDispatcher at 26310992 for <Subprocess at 26313168 with name produce-unicode-error in state STARTING> (stdout)>
```

mnaberez:
From the traceback, the source of the exception is the bytes/string conversion code added for Python 3 compatibility.  This should only happen on the master branch.  A release version of Supervisor like 3.1.3 won't have this problem.

Lucretiel:
It looks like we're assuming UTF-8 for all data coming off of the child process:

https://github.com/Supervisor/supervisor/blob/7dcbc2c2adc5cf2edc42799cf1c26d2e865ed021/supervisor/compat.py#L22

https://github.com/Supervisor/supervisor/blob/669ec850c9c1bb715092b8cd2c412a305b3fc735/supervisor/options.py#L1507

Would this be worth making configurable in the program options? Something like:

    [program:emit-raw]
    command: echo -e "\x88"
    stdout-encoding: latin-1

The alternative would be to alter the supervisor internals to pass only `bytes` around. This would be a pretty drastic change, and would really only defer the problem, as many parts of supervisor (event handling, `logging` logs) require text strings.

regebro:
[...]

PR #868, opened 2016-12-26
Fix buffer type in dispatcher

evanunderscore:
This fixes the error seen in #663, #835 and #836, but seems to expose further errors.

### DIFF
--- a/supervisor/dispatchers.py
+++ b/supervisor/dispatchers.py
@@ -465,7 +465,7 @@ class PInputDispatcher(PDispatcher):
 
     def __init__(self, process, channel, fd):
         PDispatcher.__init__(self, process, channel, fd)
-        self.input_buffer = ''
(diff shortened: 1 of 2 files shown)


mnaberez:
> but seems to expose further errors.

Could you elaborate?

evanunderscore:
Running up the example in #836 with this patch applied gives me this:

```
$ python3 -m supervisor.supervisorctl fg cat
==> Press Ctrl-C to exit <==
hi
b'hhttp://127.0.0.1:9001/logtail/cat/stdout Cannot connect, error: <class 'ValueError'> (invalid literal for int() with base 16: "i\\n'")
```

(Note that it doesn't crash.)

Sending trying to send some Unicode characters gives no response.

evanunderscore:
Superseded by ongoing work by @vsajip (see #638, #109)

Issue #813, opened 2016-08-23
minfds docs: child processes?


meshy:
Having read the [docs for the `minfds` setting](http://supervisord.org/configuration.html#supervisord-section-values), I'm unsure if the FD limit is applied to child processes. Is it?

I'd be happy to attempt to update the docs on this `:)`.

mnaberez:
`supervisord` calls `setrlimit` once when it starts (from [here](https://github.com/Supervisor/supervisor/blob/3.3.1/supervisor/supervisord.py#L70) using [this](https://github.com/Supervisor/supervisor/blob/3.3.1/supervisor/options.py#L1327)).  It does not call `setrlimit` again before fork/exec, so the limits will be inherited by the subprocess.

meshy:
Thank you for the clarification. Do you agree with me that the docs are unclear on this point and, in this regard, could be improved?

mnaberez:
Docs are in [configuration.rst](https://github.com/Supervisor/supervisor/blob/master/docs/configuration.rst), feel free to submit PRs.

mnaberez:
To clarify: I'm not sure if it's an issue or not because this is the first time this question has been asked.  It certainly wouldn't hurt to add more clarification to the docs. You had offered above to help, and a PR would be welcome.  Thanks.

PR #872, opened 2016-12-30
Explicitly document minfds limit inheritance

meshy:
Child processes will inherit the file descriptor limits, so this amends the documentation to make this absolutely clear, as discussed in #813.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -315,7 +315,8 @@ follows.
   is run as root.  supervisord uses file descriptors liberally, and will
   enter a failure mode when one cannot be obtained from the OS, so it's
   useful to be able to specify a minimum value to ensure it doesn't run out
-  of them during execution. This option is particularly useful on Solaris,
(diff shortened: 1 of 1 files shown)


docs/configuration.rst
```diff
@@ -315,7 +315,8 @@ follows.
   is run as root.  supervisord uses file descriptors liberally, and will
   enter a failure mode when one cannot be obtained from the OS, so it's
   useful to be able to specify a minimum value to ensure it doesn't run out
-  of them during execution. This option is particularly useful on Solaris,
+  of them during execution.  These limits will be inherited by the managed
+  subprocesses.  This option is particularly useful on Solaris,
   which has a low per-process fd limit by default.
```
meshy: I've just realised that I've not been completely consistent with line wrapping here. Would you like me to wrap this final line in the paragraph?

mnaberez: No problem on the line wrapping.

mnaberez:
Thanks!

PR #880, opened 2017-01-17
Fix failure to parse environment with quoted empty strings.

### LINKED ISSUE #873: Failure to parse environment with quoted empty strings. (Issue #328)
commit aa7d40e2949043ece15f972a is incorrect...

 The problem is that the test in the above commit tests the value 'foo=""' -- that works.
However, 'foo="",bar=a' will fail.  This seems to be related to the use of "posix=True" with shlex.

#>>> supervisor.datatypes.dict_of_key_value_pairs('foo="",c=d')
  File "<stdin>", line 1, in <module>
  File "/home/mostrows/workspace/cmf.build/agents/cmf/ext-py/supervisor-4.0.0.dev0/supervisor/datatypes.py", line 96, in dict_of_key_value_pairs
    "Unexpected end of key/value pairs in value '%s'" % arg)
ValueError: Unexpected end of key/value pairs in value 'foo="",c=d'
 #>>> supervisor.datatypes.dict_of_key_value_pairs("foo='',c=d")
  File "<stdin>", line 1, in <module>
  File "/home/mostrows/workspace/cmf.build/agents/cmf/ext-py/supervisor-4.0.0.dev0/supervisor/datatypes.py", line 96, in dict_of_key_value_pairs
    "Unexpected end of key/value pairs in value '%s'" % arg)
ValueError: Unexpected end of key/value pairs in value 'foo='',c=d'
#>>> supervisor.datatypes.dict_of_key_value_pairs("foo=''")
{'foo': ''}
#>>> supervisor.datatypes.dict_of_key_value_pairs('foo=""')
{'foo': ''}
#>>>

coldnight:
The `shlex` in standard library has a bug before Python 2.7.13 and 3.6.0. Try keep an newest version in local to solve this problem.

The bug detail of shlex: http://bugs.python.org/issue21999

This pull request fixes #873.

### DIFF
--- a/supervisor/_shlex2.py
+++ b/supervisor/_shlex2.py
@@ -1,0 +1,292 @@
+# -*- coding: iso-8859-1 -*-
+"""A lexical analyzer class for simple shell-like syntaxes."""
+
+# Module and documentation by Eric S. Raymond, 21 Dec 1998
+# Input stacking and error message cleanup added by ESR, March 2000
+# push_source() and pop_source() made explicit by ESR, January 2001.
+# Posix compliance, split(), string arguments, and
+# iterator interface by Gustavo Niemeyer, April 2003.
+
+import os.path
+import sys
+from collections import deque
+
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from StringIO import StringIO
+
+__all__ = ["shlex", "split"]
+
+class shlex:
+    "A lexical analyzer class for simple shell-like syntaxes."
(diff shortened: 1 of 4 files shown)


mnaberez:
Thanks for your work on this.  I'd rather not bundle two complete shlex implementations inside Supervisor just to fix #328, so I've committed fa7e3f56c8e4fd1e6e19070a0bcffc0bf5ab3192 instead.  I'd still like to fix #328, so hopefully we can find a simpler solution.

Issue #858, opened 2016-12-03
supervisor run error After supervisorctl reload command execute twice


zhuqiuyang:
why supervisorctl reload cmd will lead supervisor report err and shut down.
i prepare use it to daemon my golang application.
![image](https://cloud.githubusercontent.com/assets/14734729/20857873/977719ea-b973-11e6-8db3-aaa69e9c58a4.png)

mnaberez:
#### Reproduce Instructions

Start `supervisord` in the foreground with this config:

```
$ cat supervisord.conf
[supervisord]

[inet_http_server]
port = 127.0.0.1:9001

[supervisorctl]
serverurl = http://127.0.0.1:9001

[rpcinterface:supervisor]
supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface

$ supervisord -n -c supervisord.conf 
2016-12-29 14:09:05,033 INFO Increased RLIMIT_NOFILE limit to 1024
2016-12-29 14:09:05,044 INFO RPC interface 'supervisor' initialized
2016-12-29 14:09:05,045 CRIT Server 'inet_http_server' running without any HTTP authentication checking
2016-12-29 14:09:05,045 INFO supervisord started with pid 22137
```

While `supervisord` is still running, in another terminal:

```
$ supervisorctl -c supervisord.conf reload
Restarted supervisord

$ supervisorctl -c supervisord.conf reload
Restarted supervisord
```

After the second `reload`, `supervisord` will crash:
```
2016-12-29 14:11:49,406 INFO RPC interface 'supervisor' initialized
2016-12-29 14:11:49,406 CRIT Server 'inet_http_server' running without any HTTP authentication checking
2016-12-29 14:11:49,406 INFO supervisord started with pid 22235
2016-12-29 14:11:50,741 INFO RPC interface 'supervisor' initialized
2016-12-29 14:11:50,741 CRIT Server 'inet_http_server' running without any HTTP authentication checking
2016-12-29 14:11:50,741 INFO supervisord started with pid 22235
Traceback (most recent call last):
  File "/usr/local/bin/supervisord", line 11, in <module>
    load_entry_point('supervisor', 'console_scripts', 'supervisord')()
  File "/path/to/supervisor/supervisord.py", line 361, in main
    go(options)
  File "/path/to/supervisor/supervisord.py", line 371, in go
    d.main()
  File "/path/to/supervisor/supervisord.py", line 82, in main
    self.run()
[...]

PR #885, opened 2017-01-29
Close pollers during cleanup

jaredsuttles:
Ensuring proper cleanup for kqueue poller seems to fix https://github.com/Supervisor/supervisor/issues/858


mnaberez:
Added tests in ea76b3e5e352a3efb516f025ddfdbaed33f896ee and changelog entry in ea76b3e5e352a3efb516f025ddfdbaed33f896ee.  I will include this patch in the next bug fix release.  Thanks again.

Issue #898, opened 2017-03-03
Add an option to require program sections to specify a non-root user


groutr:
We are looking at supervisord to manage many user-level services in an enterprise setting.  Running supervisord as root is attractive, but we need to be able to guarantee every process supervisor executes drops root privileges.

AFAIK, supervisord doesn't make any such guarantee.  We're interested in adding an option to supervisord (like `--no_root_processes` or similar) that will function in one of the following two ways:
1. enforce each `[program:x]` in supervisor.conf (especially any included files) to have a user key defined and it cannot be root.
2. for each file that is included, override/set the uid of each `[program:x]` section to the owner of the included file.

If this is already possible, I'd love learn about how configure supervisor this way.  Otherwise, I'm willing to make the changes in a PR, but I think I might need a little guidance understanding the internals of supervisor.

mnaberez:
> We are looking at supervisord to manage many user-level services in an enterprise setting. Running supervisord as root is attractive, but we need to be able to guarantee every process supervisor executes drops root privileges.

Run supervisord itself as non-root or choose another program instead of Supervisor to enforce enterprise security.  Supervisor just wasn't designed for enforcement of enterprise security as its main function and as such it isn't as paranoid as other programs such as daemontools.  While I'm not aware of an particular vulnerabilities in Supervisor at this time, I don't think Supervisor is a suitable tool for your application.  It was just never designed for the use case you are describing.  I'm not keen on adding config file options that might appear like it was.  Use a different program that was designed from the ground up to enforce enterprise security.

groutr:
This is unfortunate.  Supervisor was the most appealing option.

Perhaps enforce was an incorrectly chosen word.  The two approaches explained above are really more of a validation of the configuration.

mnaberez:
> I'm not trying to say that supervisor be responsible for implementing/enforcing enterprise-grade security.

OK.  Be aware of that supervisord switches the user by calling `setuid()` only, so it doesn't know about PAM.  For example, it is possible that after supervisord drops privileges, that user ends up with higher resource limits than it would have otherwise, because it inherits the limits from supervisord.  So if you're security conscious, I maintain you should run supervisord as non-root or look for another option.

> AFAIK, supervisord doesn't make any such guarantee. We're interested in adding an option to supervisord (like --no_root_processes or similar) that will function in one of the following two ways:
>
>  1.  enforce each [program:x] in supervisor.conf (especially any included files) to have a user key defined and it cannot be root.

[...]

PR #899, opened 2017-03-04
Add an option to require program sections to specify a non-root user

groutr:
My initial attempt for #898.  It seems to work as I imagined during my very limited testing.

I believe this is the only change needed.  I'm open to comments or suggestions for improvements.


groutr:
Maybe a better option name would be `no_root_uid`?

mnaberez:
Please see response in https://github.com/Supervisor/supervisor/issues/898#issuecomment-284262846.  If you need this feature, please apply this patch to your own copy of supervisord.

PR #901, opened 2017-03-07
added py36

cclauss:


cclauss:
When do we get a supervisor release to PyPI?  It would be nice to see supervisor come off the red list.  https://python3wos.mybluemix.net

mnaberez:
> When do we get a supervisor release to PyPI? 

Stable (3.x) versions of Supervisor are released to PyPI fairly regularly.  The next one should be within a few weeks and will fix some bugs.  

> It would be nice to see supervisor come off the red list.

The master branch will become the next major version (4.0) and is intended to support Python 3, but as the [README](https://github.com/Supervisor/supervisor/blob/23925d017f8eccafb1be57c509a07df75490c83d/README.rst) says, Python 3 support doesn't work yet.  Those issues need to get resolved before a release could be made.  The tests pass but the test coverage apparently isn't very good.

cclauss:
Understood.  Thanks.  Supervisor is the most downloaded PyPI package on the red list.

PR #875, opened 2017-01-11
Fix 100% CPU usage

jaredsuttles:
I ran into this bug using the web UI on macOS. It seems that https://github.com/Supervisor/supervisor/pull/589 was on the right track but has unnecessarily strict rules to unregister a file descriptor. Only need to check if the writers are not writeable anymore - readers not readable.

PR #903, opened 2017-03-10
Clear writable sockets after the first event

### LINKED ISSUE #807: 100% CPU usage (maybe caused by new poll implementation?)
We were using supervisor 3.3.0 on Ubuntu 14.04 LTS.

Recently on some of our nodes on AWS, we spot very high cpu usage about supervisord, to get around that, we have to reload it, but it may happen again in one day.

Reading from `strace`, we spot there are excessive calls to both 'gettimeofday' and 'poll', so after that, we have to choose to downgrade supervisor to 3.2.3.

I see there was #581, but I think it's irrelevant here, our wild guess is it just caused by the new poll implementation introduced in 3.3.0 (and maybe caused by simultaneous log outputs?)...

Thanks in advance!

mhariri:
This fixes #807

### DIFF
--- a/supervisor/poller.py
+++ b/supervisor/poller.py
@@ -140,7 +140,7 @@ class KQueuePoller(BasePoller):
     def register_writable(self, fd):
         self.writables.add(fd)
         kevent = select.kevent(fd, filter=select.KQ_FILTER_WRITE,
-                               flags=select.KQ_EV_ADD)
+                               flags=select.KQ_EV_ADD | select.KQ_EV_CLEAR)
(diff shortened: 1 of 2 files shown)


mnaberez:
@jaredsuttles @mhariri This patch #903 and #875 both try to fix the 100% CPU usage issue reported in #807, but they are completely different patches.  Can you please have a discussion on why you think one approach or the other should be used, and see if you can come to a consensus on the best way to fix #807?  Thanks both for your help.

mhariri:
Closing in favour of #875

PR #904, opened 2017-03-11
Dependson 122

lx0079:


mnaberez:
This was discussed with the author in another ticket.  You should let him submit his patch himself when/if he thinks it's ready.

PR #919, opened 2017-04-11
Add a couple of common python dirs to .gitignore

luto:
This adds the `venv` (a common dir for virtual envs) and `.cache` (used by pip while building) directories to `.gitignore`.


luto:
Thanks!

PR #928, opened 2017-04-26
Keep sys.stdout/stderr open for debuging unexpected exit

tyronecai:
The AssertionError raised by _assertInState() is only logged to stdout/stderr.
Closing stdout/stderr make it really hard to find problems.


mnaberez:
This code is part of daemonization so the descriptors should be closed ([explanation](https://github.com/Supervisor/supervisor/blob/30232771b4ddf3b9c742e01a23d6c9e08cfd998c/supervisor/options.py#L1115)).

Issue #929, opened 2017-04-27
/logtail/<program> returns a 404 when inet_http_server run behind a reverse proxy, via a sub-path


kartick:
1. Setup, and run nginx as reverse proxy, proxying to http://127.0.0.1:9001 and responding to requests coming on `/supervisor` (see attached config).
2. Fire up the browser and access http://127.0.0.1/supervisor (this will show the web UI correctly), links like stop, restart, work
3. Click the "Tail -f" link for any supervisor "program" listed on the UI -> Returns a 404

[default.txt](https://github.com/Supervisor/supervisor/files/960151/default.txt)

mnaberez:
Duplicate of #29.

kartick:
@mnaberez I see there are several related issues to this one (#929). Any idea when the fix for this issue will be incorporated?

PR #930, opened 2017-04-27
match paths when supervisor is run behind a reverse proxy

kartick:
Fix for https://github.com/Supervisor/supervisor/issues/929


mnaberez:
Closing in favor of #593 as the approach here would probably lead to false matches for some paths.

Issue #242, opened 2013-05-31
Feature request: include environment variables from external file


mkotsalainen:
I would like to keep my environment variables out the the supervisord.conf file for security reasons (I keep the supervisord.conf under version control. I've looked at the documentation and I don't see any way of doing this with supervisor. Am I missing something?

tuuling:
I found the init script at: /etc/default/supervisor

robolivable:
Use case:
A production server with 10 or so processes configured under supervisor. These processes rely on IO from a remote database to complete transactions for users. Currently, any changes to environment variables outside of the config file requires restarting the service, which could interrupt any running processes that are in the middle of transactions.

My question in this case would be, would an "external environment variable file" feature make sense here, so that variables can be defined and reread within supervisor, or is this an issue of how the child process chooses how to terminate when sent the signal?

odedfos:
As suggested above, a simple workaround would be to wrap the target command with another script that will load the environment variables from a file.

You could have a local (or source controlled) script such as:

run_with_env.sh
```
#!/bin/bash

ENV_FILE="$1"
CMD=${@:2}

set -o allexport
source $ENV_FILE
set +o allexport

$CMD
```

And then have your supervisor command be something like this:

`command=run_with_env.sh envfile some-program param1 param2`

inspired by this answer on SO [30969768/1629696](http://stackoverflow.com/a/30969768/1629696)

jonathan-golorry:
The workaround script is nice, but this still feels like something that should just be built in. Its surprising this is still coming up 4 years later.

iongion:
The workaround does not work if you need to launch a new process, processes launched from the bash file will not be killed.
I've expected the `environment=` config option to also support files, not only K=V that are not so flexible.
Currently I am stuck as I cannot duplicate all the 60+ environment variables for the hook file, it simply does not make any sense.

jonathan-golorry:
@iongion 
Have you tried `killasgroup=true` and `stopasgroup=true`?

apollo13:
@mnaberez Is there any chance that you would accept a PR to add `environment_file`? Or is there any chance for a plugin interface which could update the config after the initial read (or something similar).

@jonathan-golorry `killasgroup/stopasgroup` have their own downsides (namely that they kill the whole group ;)). A better workaround would probably be to use the bash `exec` builtin which would replace the bash process with the new program.

mnaberez:
[...]

PR #934, opened 2017-05-12
support env_file directive to add environment to subprocess by file

nhamlh:
system unit file has a EnvironmentFile directive which is useful sometime. So I thought that it would be great if supervisord has something similar.


mnaberez:
Comments in #242.

PR #952, opened 2017-06-16
supervisord to syslog

Yanpas:
Now option
```
[supervisord]
logfile=syslog
```
lets you redirect supervisord log to syslog


mnaberez:
Duplicate of #162.

xiva-wgt:
Docker container set log to nsyslog on host. Only supervisord write log files into container.
Need log to syslog without filehandler.

Issue #693, opened 2015-11-18
CRIT Set uid to user in logs


ngton:
This shouldn't be a CRIT level message, it's INFO at most.

As per the documentation:

user

If supervisord is run as the root user, switch users to this UNIX user account before doing any meaningful processing. This value has no effect if supervisord is not run as root.

A CRIT message should not be generated for documented functionality that is working as designed.

mnaberez:
I think logging this at the `INFO` level would be fine.  `Set uid to 502` is a success message and similar success messages like increasing rlimits are logged at `INFO` already.

lxlang:
:+1:

enumag:
How can I install supervisor 3.3.4 on Alpine (there is a package only for 3.3.3)? Or how can I suppress this CRIT error until a newer package is available?

PR #968, opened 2017-08-02
Make "Set uid to user" message informal

cweiske:
Was CRITical before, which is not correct.

Resolves: https://github.com/Supervisor/supervisor/issues/693

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1278,7 +1278,7 @@ class ServerOptions(Options):
             return None
         msg = self.dropPrivileges(self.uid)
         if msg is None:
-            return 'Set uid to user %s' % self.uid
+            self.logger.info('Set uid to user %s' % self.uid)
         return msg
 
     def dropPrivileges(self, user):


supervisor/options.py
```diff
@@ -1278,7 +1278,7 @@ def set_uid(self):
             return None
         msg = self.dropPrivileges(self.uid)
         if msg is None:
-            return 'Set uid to user %s' % self.uid
+            self.logger.info('Set uid to user %s' % self.uid)
```
mnaberez: `self.logger` can't be used in `set_uid()` because logging hasn't been set up yet when this function is called.  That is why `set_uid()` and `dropPrivileges()` both return the error message in a string.  Below is what happens when this patch is run.

```
# whoami 
root 

# cat supervisord.conf 
[supervisord]
user = supervisor ;valid user on this system

# supervisord -n -c supervisord.conf
  File "/usr/local/bin/supervisord", line 11, in <module>
    load_entry_point('supervisor', 'console_scripts', 'supervisord')()
  File "/path/to/supervisor/supervisor/supervisord.py", line 359, in main
    go(options)
  File "/path/to/supervisor/supervisor/supervisord.py", line 369, in go
    d.main()
  File "/path/to/supervisor/supervisor/supervisord.py", line 64, in main
    setuid_msg = self.options.set_uid()
  File "/path/to/supervisor/supervisor/options.py", line 1281, in set_uid
    self.logger.info('Set uid to user %s' % self.uid)
AttributeError: ServerOptions instance has no attribute 'logger'
```

mnaberez:
Fixed in e536fb1f03697135cd3be961aa5c5c715aab6423.

PR #974, opened 2017-08-04
Decode `chars` before appending to input buffer

abn:
Resolves: #663 

This PR ensures that `decode` is used before appending `chars` to the `input_buffer` in `supervisor.process.SubProcess.write()` to ensure compatibility with Python 2 and 3.

This does not attempt to resolve the full port to handle Python 3 byte/string issues, but rather focuses on a simple fix that is reusable without change when all strings are converted to byte-strings.

Tested locally with py27 and py36.


robdoesstuff:
What needs to happen to get this merged into master?

ibnbay00:
please merge for support python 3 :)

PR #985, opened 2017-08-26
supervisorctl output with color when it is a terminal

FAKERINHEART:
When it is a terminal to output for supervisorctl command, it may be better with colorful output.


mnaberez:
Thanks for this pull request but there's a more complete implementation in #530.  I'm closing this pull request in favor of that one.

PR #1006, opened 2017-10-20
two patches

anqfsh:
1. fixed the issue: the message url will redirect to root. this will be a problem when supervisor is deployed into a subdir instead of the root of the website.

2. add a new directive I really need: autodir
can use it in the conf file like this:
autodir=%(ENV_XXXX)s/tmp/%(program_name)s,%(ENV_XXXX)s/logs/%(program_name)s

will create the separate logs and tmp directory for every program

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -917,7 +917,6 @@ class ServerOptions(Options):
             for one_dir in autodir:
                 if not os.path.exists(one_dir):
                     os.makedirs(one_dir)
-                    print 'autodir %s created' % one_dir
             #end of new directive: autodir
 
             for k in ('stdout', 'stderr'):
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -895,6 +895,10 @@ class ServerOptions(Options):
             raise ValueError(
                 "Cannot set stopasgroup=true and killasgroup=false"
                 )
+        #begin of new directive: autodir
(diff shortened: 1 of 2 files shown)


mnaberez:
> fixed the issue: the message url will redirect to root. this will be a problem when supervisor is deployed into a subdir instead of the root of the website.

This will be fixed by #593.

> add a new directive I really need: autodir

Sorry, but I don't think we will add a feature that creates arbitrary directories from a comma-separated list in the config file.  

If you need to create some directories, or perform any other kind of initialization before your program is run, you can do something like the following:

Original:
```
[program:myprogram]
command = myprogram
```

Change it to:
```
[program:myprogram]
command = bash -c start_myprogram.sh
```

Contents of `start_myprogram.sh`:
```
#!/bin/bash 
mkdir dir1
mkdir dir2 
exec myprogram
```

It will create the directories and then the `exec` will cause the process to be replaced by `myprogram`, so the end effect is the same as if `myprogram` was started by `supervisord` directly.

FAKERINHEART:
I think it is necessary to add create log dir by **supervisor** instead of every sub-process, for these log are written into by **supervisor**, not the sub-process. And it is necessary for supervisor to have enough permission to access them. In addition, the log dir of supervisord itself is also need to be created.

PR #1020, opened 2017-11-19
WIP: Add possibility to read config from remote file

tiagocoutinho:
I think it could be interesting if both `supervisord` and `supervisorctl` are able to read configuration from a remote location by extending the config parameter to allow a url:

```bash
$ supervisord -c http://git.acme.com/coyote/master/supervisor.ini
```
Tasks:
* [ ] supervisord:
  * [x] basic remote file (no `%(here)s`; no `[include]`)
  * [ ] handle `%(here)s`
  * [ ] handle `[include]`
  * [ ] tests
  * [ ] doc
* [ ] supervisorctl 
  * [x] support remote file
  * [x] properly handle `%(here)`
  * [ ] tests
  * [ ] doc

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -22,6 +22,8 @@ from supervisor.compat import as_bytes, as_string
 from supervisor.compat import xmlrpclib
 from supervisor.compat import StringIO
 from supervisor.compat import basestring
+from supervisor.compat import urllib
+from supervisor.compat import urlparse
 
 from supervisor.medusa import asyncore_25 as asyncore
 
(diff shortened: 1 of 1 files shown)


mnaberez:
I will not merge a feature that loads a config file from an arbitrary remote URL.  I do not see a need for such a feature, but much more than that, I do not want to introduce all the possible security implications.  

Supervisor has a plugin system and it is possible for a plugin to modify `supervisord`'s configuration (e.g. the [twiddler](https://github.com/mnaberez/supervisor_twiddler) plugin does this).  A plugin could be written to load configuration from a URL, or you can fork `supervisord`, or you could use various solutions that don't require writing code like NFS or FUSE filesystems.

tiagocoutinho:
Ok , fair enough.
I will have a look at [twiddler](https://github.com/mnaberez/supervisor_twiddler). Thanks for the suggestions.

PR #1034, opened 2017-12-30
Add Travis badge

szepeviktor:
@mnaberez

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,9 @@
 Supervisor
 ==========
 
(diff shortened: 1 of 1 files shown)


mnaberez:
Thanks, but I don't think this is needed.  GitHub shows the CI status on the commits listing and before pull requests are merged, which is enough.  The README file is also included verbatim on the [PyPI page](https://pypi.python.org/pypi/supervisor) for stables releases, where showing the current Travis build status probably isn't desirable.

PR #1064, opened 2018-02-24
try making log dir when log dir did not exist 

FAKERINHEART:


FAKERINHEART:
@mnaberez 
I think it is necessary to add create log dir by **supervisor** instead of every sub-process, for these log are written into by **supervisor**, not the sub-process. And it is necessary for **supervisor** to have enough permission to access them. In addition, the log dir of supervisord itself is also need to be created.

mnaberez:
No, it's not necessary for supervisord to silently create directories that do not exist.  Create the directories before starting supervisord, or before starting the subprocesses.  It has worked this way for over a decade and it's been fine, and most other software works this way too.

PR #1068, opened 2018-03-03
replace hard link with symlink

Hax4us:
as we know android does not support hard links so symlinks will be better


FAKERINHEART:
The replace of hard link with symlink directly is not correct, because supervisord will remove the tempsocket after the os.link(tempname, socketname), and use the socketname link to do unix socket communication.

Hax4us:
Would you like to commit some changes as you suggested :) I will also look

FAKERINHEART:
#1069 
@Hax4us

Hax4us:
Nice

Issue #123, opened 2012-05-31
Unable to specify log file owner or permissions


lra:
When I start supervisor, the log files of my programs have 0600 permissions, while the supervisor.log is 0644.

```
ubuntu@sentry:/var/log/supervisor$ ls -l /var/log/supervisor/
total 20
-rw------- 1 root root 7975 May 31 18:54 sentry_celeryd-stderr---supervisor-1gPFQa.log
-rw------- 1 root root    0 May 31 18:16 sentry_celeryd-stdout---supervisor-dZn9PW.log
-rw------- 1 root root 4561 May 31 19:02 sentry-stderr---supervisor-GUllAv.log
-rw------- 1 root root    0 May 31 18:16 sentry-stdout---supervisor-8HXvhm.log
-rw------- 1 root root    0 May 31 18:16 sentry_udp-stderr---supervisor-uAlO19.log
-rw------- 1 root root    0 May 31 18:16 sentry_udp-stdout---supervisor-PhobKI.log
-rw-r--r-- 1 root root 4039 May 31 18:16 supervisord.log
```

How can I specify another umask that will be applied to program's logs ?

yeago:
oh supervisord.

jonnywilliamson:
Just another user looking for exactly the same thing!

Erve1879:
Setting ``user=<user>`` on the ``supervisord`` program works for me... All logs are chown'd to that user by supervisord.

breerly:
+1 please - this is a pretty big deal

gkertai:
We were also struggling with this. The workaround we came up with involves pre-creating all log files as part of our deployment process before `supervisord` restarts. Basically we run a `grep | awk | xargs (touch ; chown; chmod)` command against the supervisor configs. This works fine as long as the logs don't rotate, which they rarely do.

A nicer solution would be appreciated though. ;)

neuroticnerd:
+1
[...]

PR #1075, opened 2018-03-10
chmod temp files using umask from ServerOptions

ericatkin:


mnaberez:
> ***No description provided.***

Please give us more to work with here.  What was the motivation for this change, did you find a problem that this is intended to fix, etc.?

JonathanGrant:
lgtm

ericatkin:
So, this isn't really a feature change, it's a bugfix. What I mean is, a use case doesn't seem needful as the use case is to create log files as documented. This change fixes that bug described in issue #123 that caused the behavior to not match the documentation.
Thanks.

mnaberez:
> This change fixes the bug described in issue #123 that caused the behavior to not match the documentation.

Thanks for adding some background on this change.  Could you please point to the specific page(s) in the documentation that are inconsistent?

mnaberez:
@ericatkin We will probably have another point release soon.  Independent of this PR or #123, we can immediately fix the [incorrect documentation you mentioned](https://github.com/Supervisor/supervisor/pull/1075#issuecomment-414737232).  I had a look and didn't find it, but there's a lot of docs so I probably just missed it.  It would be great if you could point out the pages that are inconsistent and we can correct them as a first step.  Thanks.

ericatkin:
@mnaberez I didn't review this issue when I created the comment referring to incorrect documentation. It appears I was wrong about that. I also can't find any documentation that explicitly states log files will be created with permissions derived from the process umask. That said, it seems to me like a reasonable assumption from the language on the [umask](http://supervisord.org/configuration.html#program-x-section-values) documentation as well as the behavior of many other CLI utilities. This PR make it behave according to that assumption. I think _not_ creating files with the process umask is the surprising bit. Either this PR (or a similar change) should be merged, or the documentation updated to describe the exceptional behavior. I of course am in favor of the first option :)

PR #1097, opened 2018-05-05
start stop cmd support wildcard *

Geew:


Geew:
It would be better to change * to %

azmeuk:
Why that?

azmeuk:
@mnaberez Sadly this plugin is not packaged in most distributions. I would really love if you could reconsider integrating this feature directly in supervisor, that would make things simplier.
The PR size is quite small, so there is not a lot of further code to maintain.

mnaberez:
The Supervisor project only publishes packages to PyPI.  The distribution packages are created by third parties not involved in the project, and they tend to lag behind the Supervisor releases by years.  It's unlikely that adding code to Supervisor will cause it to appear in a distribution any time soon.  It might be easier to ask your distribution to package the plugin.

PR #1100, opened 2018-05-15
Add vim modeline to sample configuration file

sbraz:
Hi, my vim install did not recognise the file syntax properly so I think that adding a modeline is a good idea, it makes it much easier to see what parts are commented out.


mnaberez:
We do not include any editor or IDE specific bits in this project, sorry.

PR #1137, opened 2018-08-21
Replace easy_install by pip installation method

sblondon:
Today, `pip` has replaced `easy_install` to install python packages.

This PR updates documentation and fixes issue #623


sblondon:
The new commit merges the two section, keeping `pip` as the first section because it's probably the most used method (with distribution packaging method).

PR #1163, opened 2018-10-22
Added support for fcgi-program in groups, fixes #148.

### LINKED ISSUE #148: Support for fcgi-programs in groups?
Supervisord fails to start when I add a fcgi-program to a group and I found the following patch adding support in the mailing list from 2009. Would it be possible to add the patch?

http://lists.supervisord.org/pipermail/supervisor-users/attachments/20090605/98b18fb8/attachment.obj

apollo13:


apollo13:
@mnaberez Anything you'd need me to do here to get this merged?

PR #1172, opened 2018-11-15
Convert tail file to utf-8

fitzg2:
Fixes a bug where if a log file has strange characters it will prompt for downloading the log file. Indefinitely downloading the log file. With this change it will convert the file to utf-8 for display.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -15,6 +15,7 @@ import glob
 import platform
 import warnings
 import fcntl
+import codecs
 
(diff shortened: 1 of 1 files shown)


fitzg2:
Wait. That conversion worked for me once and now it keeps downloading a log file.

mnaberez:
Closing due to inactivity.  If you are still experiencing a problem, please submit an issue with instructions for how to reproduce it.

PR #1173, opened 2018-11-18
Web UI: Renamed "Tail -f" to stdout and added "stderr" to get tail -f stderr

ouroboroscoding:
I have been using supervisor for years and it has been forever frustrating that there is only the "Tail -f" option when often stderr is the thing I really want to see, in my fork I have replaced "Tail -f" with "stdout" and added an equivalent "stderr" which calls "'logtail/%s/stderr' % processname". I would be forever grateful if you would consider making this change to supervisor.

### DIFF
--- a/supervisor/web.py
+++ b/supervisor/web.py
@@ -260,18 +260,23 @@ class StatusView(MeldView):
         'href':'index.html?processname=%s&amp;action=clearlog' % processname,
         'target':None,
         }
-        tailf = {
-        'name':'Tail -f',
+        stdout = {
+        'name':'stdout',
         'href':'logtail/%s' % processname,
         'target':'_blank'
         }
+        stderr = {
+        'name':'stderr',
+        'href':'logtail/%s/stderr' % processname,
+        'target':'_blank'
+        }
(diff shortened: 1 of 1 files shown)


jouve:
@mnaberez what is blocking this PR ?

djchou:
Please merge this -- its useful.

ouroboroscoding:
Can you at least explain why this was rejected? It's clearly helpful to some people and takes nothing away from the existing software.

mnaberez:
> Can you at least explain why this was rejected? It's clearly helpful to some people and takes nothing away from the existing software.

Please see the link above to a88a19315e7795e2ed8043ab9c1862b4646cf18e.  I didn't reject it, I just did it slightly differently.  This patch adds a new stderr link but it also removes `tail -f` from the link text.  I didn't want to remove `tail -f` so that it remains clear to users that the function is `tail -f`.  I credited you for the original patch in the changelog.  Thank you.

ouroboroscoding:
Ah, ok, I missed that. Thanks so much, this is super helpful :)

PR #1185, opened 2018-12-16
xmlrpc: properly close connection on SupervisorTransport.close

### LINKED ISSUE #1184: SupervisorTransport does not close connections
python3's unittest module report unclosed files and sockets, with this minimal test, it complain that connection is not closed ( this is on python 3.5.3 ):

```
$ python3 -m unittest close_repro.py
/usr/lib/python3.5/unittest/case.py:601: ResourceWarning: unclosed <socket.socket fd=4, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/tmp/super.sock.7689>
  testMethod()
.
----------------------------------------------------------------------
Ran 1 test in 0.196s

OK
``` 

```python
import unittest
import xmlrpc.client
import tempfile
import subprocess

import supervisor.xmlrpc

class Repro(unittest.TestCase):
  def test_repro(self):
    with open('super.conf', 'w') as f:
      f.write('''
[unix_http_server]
file = /tmp/super.sock

[supervisorctl]
serverurl = unix:///tmp/super.sock

[supervisord]
pidfile = /tmp/super.pid

[rpcinterface:supervisor]
      ''')
      f.flush()
      subprocess.check_call(['supervisord', '-c', f.name])

      with xmlrpc.client.ServerProxy(
        'http://localhost',
        transport=supervisor.xmlrpc.SupervisorTransport(
           None,
           None,
           serverurl="unix:///tmp/super.sock")) as server:
        server.supervisor.shutdown()
```

perrinjerome:
Method on parent class ( xmlrpclib.Transport.close ) expect connection to be
self._connection[1], but SupervisorTransport use self.connection for the
socket. Implement SupervisorTransport.close which uses self.connection.

Fixes #1184

### DIFF
--- a/supervisor/tests/test_xmlrpc.py
+++ b/supervisor/tests/test_xmlrpc.py
@@ -508,6 +508,17 @@ class SupervisorTransportTests(unittest.TestCase):
         self.assertEqual(dummy_conn.requestargs[3]['Accept'], 'text/xml')
         self.assertEqual(result, ('South Dakota',))
 
+    def test_close(self):
+        transport = self._makeOne('user', 'pass', 'http://127.0.0.1/')
+        dummy_conn = DummyConnection(200, '''<?xml version="1.0"?>
+        <methodResponse><params/></methodResponse>''')
+        def getconn():
+            return dummy_conn
+        transport._get_connection = getconn
+        transport.request('localhost', '/', '')
+        transport.close()
+        self.assertTrue(dummy_conn.closed)
+
 class TestDeferredXMLRPCResponse(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


perrinjerome:
for reference, `ServerProxy.close` calls `Transport.close`, which implementation is
 * on python2.7 https://github.com/python/cpython/blob/2.7/Lib/xmlrpclib.py#L1396-L1400
 * on pyton 3.7 https://github.com/python/cpython/blob/3.7/Lib/xmlrpc/client.py#L1251-L1255 ( it's same )

PR #1220, opened 2019-04-01
Add a new CLI option called "--strict_config_parsing_off"

ethanhann:
This change adds backward compatibility for config file originally designed to work without strict parsing. For Py2, the config parser has strict parsing turned OFF by default. Strict parsing is turned ON by default for Py3.

**Edit**
Description of `strict` option from the Python docs:
"When set to True, the parser will not allow for any section or option duplicates while reading from a single source (using read_file(), read_string() or read_dict())."

Practically speaking, if you are upgrading from Py2 to Py3 you may have supervisor config files that have duplicate options (e.g. "environment=" multiple times for multiple environment variables). This change allows existing config files to be used to ease the transition to supervisor on Py3.

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -447,6 +447,7 @@ class ServerOptions(Options):
                  "t", "strip_ansi", flag=1, default=0)
         self.add("profile_options", "supervisord.profile_options",
                  "", "profile_options=", profile_options, default=None)
+        self.add(None, None, "", "strict_config_parsing_off")
         self.pidhistory = {}
         self.process_group_configs = []
         self.parse_criticals = []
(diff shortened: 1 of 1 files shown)


mnaberez:
What are the practical effects of strict parsing on the Supervisor config file?  Can you please explain why someone would want or need to turn it off?

mnaberez:
#### Reproduce Instructions

Create `supervisord.conf`:

```
[supervisord]

[program:duplicate]
command=/bin/cat

[program:duplicate]
command=/bin/cat
```

Run:
```
$ supervisord -c supervisord.conf
```

On Python 2, no exception is raised.  On Python 3, ``configparser.DuplicateSectionError`` is raised.

mnaberez:
Thanks for reporting this issue and your pull request.  Although I prefer the new strict behavior, I have made it always non-strict in 779f4ab7bdfd4df5ffc6f236c26311d3732525c3.  In order to make  config files compatible between Supervisor versions, and to ease migration, the config file parsing should be the same on Python 2 and 3 if possible (this is what we did with `inline_comment_prefixes` as well).  I would rather not add new command line args that users have to discover.  Perhaps in a future version we will make the config file parse as strict, but we will do it without args and make it work the same everywhere Supervisor runs.

ethanhann:
Thanks for the quick handling of this. Much appreciated.

PR #1237, opened 2019-05-01
Added "STDOUT" value for "stdout_logfile" option

babinskiy:
There is issue with using `supervisor` with `systemd`.
The `systemd` redirects stdout/stderr for runned processes using sockets.
But `supervisor` for logging child's stdout/stderr uses an system call `open` that does not know how to work with sockets ([similar situation described here](https://bugzilla.redhat.com/show_bug.cgi?id=1212756#c4)), so when you specify 
```
stdout_logfile=/dev/stdout
```
in configuration you will get error like  
```
INFO spawnerr: unknown error making dispatchers for 'app1': ENXIO
```
This PR adds the ability to redirect the output of the program through `StreamHandler` instead `FileHandler`, allowing you forward logs directly to `journald`.

PS: Sorry for my English. I'm not a fluent speaker

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -869,9 +869,11 @@ where specified.
   true, also place stderr output in this file).  If ``stdout_logfile``
   is unset or set to ``AUTO``, supervisor will automatically choose a
   file location.  If this is set to ``NONE``, supervisord will create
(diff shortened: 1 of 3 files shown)


vsajip:
Whoops, that should have been #1327

mnaberez:
Duplicate of #849

PR #1244, opened 2019-05-28
Added warning message about encodings to "tail" command as well as "t…

vsajip:
…ail -f".

This involved a bit of refactoring of common code. Also added a test.


mnaberez:
Thanks!

PR #1252, opened 2019-06-12
Fix tail bug

aruseni:
The bug was that when the channel was explicitly specified, it always
evaluated to stderr, i.e. supervisorctl tail foo stdout and
supervisorctl tail foo stderr were both showing stderr.
That is because we should not use is for such comparisons:
these two are different objects.

If curious, read more here: https://stackoverflow.com/questions/1504717/

This just changes the tail command to use == for comparing the two objects.

### DIFF
--- a/supervisor/supervisorctl.py
+++ b/supervisor/supervisorctl.py
@@ -538,7 +538,7 @@ class DefaultControllerPlugin(ControllerPluginBase):
         else:
             check_encoding(self.ctl)
             try:
-                if channel is 'stdout':
+                if channel == 'stdout':
                     output = supervisor.readProcessStdoutLog(name,
                                                              -bytes, 0)
                 else: # if channel is 'stderr'


mnaberez:
> The bug was that when the channel was explicitly specified, it always evaluated to stderr

Here's what I tried to reproduce this.  

Python script to print on `stdout` and `stderr`:

```
import sys
import time

while True:
  sys.stderr.write('stderr ')
  sys.stderr.flush()
  sys.stdout.write('stdout ')
  sys.stdout.flush()
  time.sleep(0.25)
```

Config file:

```
[supervisord]
nodaemon = true

[inet_http_server]
port = 127.0.0.1:9001

[supervisorctl]
serverurl = http://127.0.0.1:9001

[rpcinterface:supervisor]

[program:print]
command = python3 print.py
startsecs = 0
autorestart = false
```

Start `supervisord` in the foreground with that config:

```
$ supervisord --version 
4.0.3

$ supervisord -n -c supervisord.conf 
2019-06-12 18:26:18,620 INFO Increased RLIMIT_NOFILE limit to 1024
2019-06-12 18:26:18,635 INFO RPC interface 'supervisor' initialized
2019-06-12 18:26:18,635 CRIT Server 'inet_http_server' running without any HTTP authentication checking
2019-06-12 18:26:18,635 INFO supervisord started with pid 1676
2019-06-12 18:26:19,643 INFO spawned: 'print' with pid 1679
2019-06-12 18:26:19,692 INFO success: print entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
```

In another terminal:

```
$ supervisorctl -c supervisord.conf
print                            RUNNING   pid 1739, uptime 0:00:19

supervisor> tail print
stdout stdout stdout stdout stdout stdout stdout stdout stdout 

supervisor> tail -f print 
==> Press Ctrl-C to exit <==
stdout stdout stdout stdout stdout stdout stdout stdout stdout ^C

supervisor> tail print stdout
stderr stderr stderr stderr stderr stderr stderr stderr stderr 

supervisor> tail -f print stdout
==> Press Ctrl-C to exit <==
stdout stdout stdout stdout stdout stdout stdout stdout stdout ^C

supervisor> tail print stderr
stderr stderr stderr stderr stderr stderr stderr stderr stderr  

supervisor> tail -f print stderr
==> Press Ctrl-C to exit <==
stderr stderr stderr stderr stderr stderr stderr stderr stderr ^C 
```

Confirmed the bug (see `tail print stdout` above).  It affects `tail` but not `tail -f`.  The `tail` code hasn't changed in a very long time.  This bug has probably existed for many Supervisor versions.  

I'll test your patch and merge it for the next release.  Thanks!

PR #1277, opened 2019-09-15
split http port properly

### LINKED ISSUE #1276: Http server errors out.
I had the supervisor server stood up, and for some reason it fails to extract the server name / http host value from the request headers. This used to work but after a fresh pip install of supervisor it started to give me this error. 
```
ERRO :Server Error: <class 'TypeError'>, unsupported operand type(s) for +: 'NoneType' and 'str': file: /usr/local/src/supervisor/supervisor/http.py line: 328
````

jouve:
fix #1276

### DIFF
--- a/supervisor/http.py
+++ b/supervisor/http.py
@@ -316,7 +316,7 @@ class deferring_http_request(http_server.http_request):
 
         if 'HTTP_HOST' in environ:
             host = environ['HTTP_HOST'].strip()
-            parsed = urlparse.urlparse(host)
+            parsed = urlparse.urlparse('//' + host)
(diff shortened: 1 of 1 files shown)


mnaberez:
I reverted the use of `urlparse()` in ff7f18169bcc8091055f61279d0a63997d594148, so this no longer merges.  The switch to `urlparse()` was originally made to silence Python 3.8 deprecation warnings.  However, as seen in ff7f18169bcc8091055f61279d0a63997d594148, there were several more uses of `urlparse()` than just the one fixed in this pull request.  Those may have different behavior and be broken also.  It seemed safer to just back out the change until they can all get audited and fixed together.

PR #1278, opened 2019-09-15
replace python logic by css

jouve:


mnaberez:
It doesn't use CSS 3 features for compatibility with more browsers.

PR #1279, opened 2019-09-18
Fix misspelling of process

theo-o:
Process is misspelled.


supervisor/supervisorctl.py
```diff
@@ -1053,7 +1053,7 @@ def _formatChanges(self, added_changed_dropped_tuple):
             for name in names:
                 self.ctl.output("%s: %s" % (name, changedict[name]))
         else:
-            self.ctl.output("No config updates to processes")
```
mnaberez: "Process" in this context refers to a program running under supervisord.  This message uses the plural form "processes" because it refers to more than one.

Issue #1282, opened 2019-09-26
Syslog support in python3 is broken


Joppie:
## Problem

When running `supervisord` under python3 with a configuration that includes the program option `stdout_syslog = true`, messages do not show up in syslog.

## Versions tested

- 4.0.4
- ff7f18169bcc8091055f61279d0a63997d594148 (latest master at the time)

## Debugging

When running `supervisord` with `--nodaemon`, exception tracebacks reveal the problem:

```
Traceback (most recent call last):
  File "/home/staley/.local/lib/python3.7/site-packages/supervisor/loggers.py", line 372, in emit
    for line in message.rstrip('\n').split('\n'):
TypeError: a bytes-like object is required, not 'str'
```

`SyslogHandler.emit` is being passed a record that contains a message objects of type `bytes` not `str`. Passing `str` objects into the `rstrip` and `split` methods of a `byte` object is not allowed.

mnaberez:
**Reproduce Instructions**

Create a config file `supervisord.conf`:

```
[supervisord]

[program:echo]
command = sh -c "echo hello"
stdout_syslog = true 
startsecs = 0
autostart = true
autorestart = false
```

Start `supervisord` in the foreground with that config:

```
$ supervisord -n -c supervisord.conf
2019-10-19 09:32:43,662 INFO supervisord started with pid 6076
2019-10-19 09:32:44,666 INFO spawned: 'echo' with pid 6079
Traceback (most recent call last):
  File ".../supervisor/loggers.py", line 372, in emit
    for line in message.rstrip('\n').split('\n'):
TypeError: a bytes-like object is required, not 'str'
2019-10-19 09:32:44,675 INFO success: echo entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2019-10-19 09:32:45,677 INFO exited: echo (exit status 0; expected)
```

The exception above from [line 372 in `loggers.py`](https://github.com/Supervisor/supervisor/blob/ff7f18169bcc8091055f61279d0a63997d594148/supervisor/loggers.py#L372) is in `SyslogHandler`.  Tested on Python 2.7 and 3.6.  The exception only occurred on Python 3.6.

mnaberez:
Fixed in #1284.

PR #1283, opened 2019-09-26
Python3 syslog fix

Joppie:


Joppie:
This pull request addresses issue #1283

Joppie:
It appears the `SyslogHandlerTests` test class is meant to verify that both ascii and unicode strings are handeled correctly. However, in `test_emit_ascii_noerror` the expression `'hello'` is used, which is not a `bytes` (ascii) type under python3. Using the expression `b'hello'` will ensure that this is an ascii string under both python2.7 and python3. Making this change will cause the test to fail under python3.

Joppie:
I'm not sold that this is the correct way to address this problem, but I thought I'd go ahead and show a potential solution. One thing of note, because the `fmt` string is in the base `Logger` class is `'%(message)s', this means the the output messages will always be unicode strings under python3. Under python2, they will be ascii strings when the input messages is ascii, and unicode strings when the input mesage is unicode. I'm not sure what the desired behavior should be.

mnaberez:
cc: @vsajip

vsajip:
Actually, the issue addressed appears ro be #1282

vsajip:
See also #1284 for the alternative approach I suggested.

Joppie:
I was unsure of the desired behavior for when messages, formats, etc should be ascii/byte strings or Unicode strings in both python2 and python3. This pull request was mainly to provide some context for issue #1282. #1284 seems like a better fix, so I'm closing this request.

vsajip:
Generally, I believe messages, formats etc. should logically be Unicode (`unicode` on Python 2 and `str` on Python 3), as they are meant to be text for human consumption. Generally, `bytes` should only be used when communicating between processes or when doing I/O (file, socket). This "purist" approach sometimes has to be modified for reasons of legacy code.

vsajip:
Just to confirm - did you try the #1284 fix in your environment? Did messages show up in syslog with that fix applied?

Joppie:
I just tested it and it does fix the issue in the environment I am using. Thanks.

PR #1284, opened 2019-09-27
Alternative approach to addressing #1282

vsajip:
All tests passing on Python2 and Python3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@
 *.swp
 *.pss
 .DS_Store
-.coverage
+.coverage*
 .eggs/
 .pytest_cache/
 .tox/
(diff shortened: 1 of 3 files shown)


vsajip:
This should address #1282 - perhaps the reporter can confirm this?

vsajip:
Hmmm - not sure yet about why the Travis jobs failed. Ran fine locally with `tox`.

vaclavbartos:
I had the same problem as discussed in #1282, #1283 and I can confirm this fix works well.

vsajip:
@mnaberez do you need more feedback from others before merging this? Both the OP of #1282 (@Joppie) and @vaclavbartos have indicated that this change worked for them.

mnaberez:
> do you need more feedback from others before merging this? 

Nope, I'm going to run it locally and then merge it.  Thank you!

mnaberez:
I tested this on both Python 2 and 3 using the [reproduce instructions](https://github.com/Supervisor/supervisor/issues/1282#issuecomment-544170227) and it worked on both.  The exception does not occur and the message is logged to syslog.  Merging.

PR #1298, opened 2019-11-25
Fix error NoneType object has no attribute 'lower'

spidgorny:
> supervisorctl tail -f worker | grep something

crashes with error ```NoneType object has no attribute 'lower' ```


mnaberez:
> crashes with error NoneType object has no attribute 'lower' 

Thanks for the patch.  What version of Python does this occur on?

spidgorny:
> Thanks for the patch. What version of Python does this occur on?

Python 2.7

mnaberez:
[Confirmed](https://travis-ci.org/Supervisor/supervisor/builds/617872227).  It only happens on 2.7 and only when the output of `supervisorctl` is piped.

Issue #1293, opened 2019-11-10
`supervisorctl tail -100 myprog stderr` command does not work


timotheecour:
(on OSX if that matters)

this works:
```
supervisorctl tail -f myprog stderr
supervisorctl tail -f myprog stdout
supervisorctl tail -100 myprog # for stdout
```

this does not work:
```
supervisorctl tail -100 myprog stderr
error: <class 'xml.parsers.expat.ExpatError'>, not well-formed (invalid token): line 5, column 25: file: /Users/timothee/homebrew/opt/python@2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/xmlrpclib.py line: 558
```

even though it should according to docs:

```
supervisorctl help tail
tail [-f] <name> [stdout|stderr] (default stdout)
Ex:
tail -f <name>          Continuous tail of named process stdout
                        Ctrl-C to exit.
tail -100 <name>        last 100 *bytes* of process stdout
tail <name> stderr      last 1600 *bytes* of process stderr
```

so there's no way to retrieve past stderr (only live stderr via tail -f) which makes it hard to diagnose issues after the fact

/cc @mnaberez  because you wrote https://github.com/Supervisor/supervisor/issues/1050#issuecomment-362091286 even this cmd does not work for me (on OSX at least)

mnaberez:
Here's what I tried to reproduce this.

Python script `spew.py` that writes to `stderr`:
```
import sys
import time

while True:
  sys.stderr.write("this is stderr ")
  sys.stderr.flush()
  time.sleep(0.1)
```

Minimal `supervisord.conf` to run it:
```
[supervisord]

[inet_http_server]
port = 127.0.0.1:9001

[supervisorctl]
serverurl = http://127.0.0.1:9001

[rpcinterface:supervisor]
supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface

[program:spew]
command = python spew.py
startsecs = 0
autorestart = false
```

Start `supervisord` in the foreground with that config:
```
$ supervisord --version
4.1.0
$ supervisord -n -c supervisord.conf
2019-11-10 10:58:10,336 INFO Increased RLIMIT_NOFILE limit to 1024
2019-11-10 10:58:10,344 INFO RPC interface 'supervisor' initialized
2019-11-10 10:58:10,344 CRIT Server 'inet_http_server' running without any HTTP authentication checking
2019-11-10 10:58:10,345 INFO supervisord started with pid 25247
[...]

PR #1306, opened 2019-12-13
fix #85 Supervisorctl tail fails on logs with ANSI codes

### LINKED ISSUE #85: Supervisorctl tail fails on logs with ANSI codes
logfile:
^[[31m^[[7memergency^[[0m: Master DOWN!

error: <class 'xml.parsers.expat.ExpatError'>, not well-formed (invalid token): line 51, column 39: file: /usr/lib/python2.6/xmlrpclib.py line: 601

timotheecour:
/cc @@mnaberez 
* fixes https://github.com/Supervisor/supervisor/issues/1293
* fixes https://github.com/Supervisor/supervisor/issues/85

this makes tail work robustly even in presence of binary non ascii data eg ansi color escape codes

EDIT

https://github.com/Supervisor/supervisor/pull/1307 actually also fixes https://github.com/Supervisor/supervisor/issues/1293 https://github.com/Supervisor/supervisor/issues/85
 via a more general solution that better interoperates with command line tools, so maybe this PR isn't essential after https://github.com/Supervisor/supervisor/pull/1307 gets merged

### DIFF
--- a/supervisor/rpcinterface.py
+++ b/supervisor/rpcinterface.py
@@ -703,8 +703,14 @@ class SupervisorNamespaceRPCInterface:
         if logfile is None or not os.path.exists(logfile):
             raise RPCError(Faults.NO_FILE, logfile)
 
+        # TODO: also provide a way to simply return logfile
+        # print(('D20191213T020824', logfile))
         try:
-            return as_string(readFile(logfile, int(offset), int(length)))
+            ret = as_string(readFile(logfile, int(offset), int(length)))
+            import base64
+            ret = base64.b64encode(ret.encode())
+            ret = ret.decode()
+            return ret
         except ValueError as inst:
             why = inst.args[0]
             raise RPCError(getattr(Faults, why))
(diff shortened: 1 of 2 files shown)


mnaberez:
We can't just change the API like this because it would break all existing clients.

aendeavor:
> We can't just change the API like this because it would break all existing clients.

All existing clients are already broken, see https://github.com/docker-mailserver/docker-mailserver/issues/1768.

mnaberez:
> All existing clients are already broken, see docker-mailserver/docker-mailserver#1768.

I read the linked ticket and my understanding is the problem reading the log fails when it contains escape sequences, which we are tracking in #85.  

This pull request would have changed the API in an incompatible way, breaking it for all clients regardless of the log content.  In addition to `supervisorctl`, there are also a lot of third party tools that use the API directly, so I don't think we can accept a solution which breaks all of them.

aendeavor:
> > All existing clients are already broken, see docker-mailserver/docker-mailserver#1768.
> 
> 
> 
> I read the linked ticket and my understanding is the problem reading the log fails when it contains escape sequences, which we are tracking in #85.  
> 
> 
> 
> This pull request would have changed the API in an incompatible way, breaking it for all clients regardless of the log content.  In addition to `supervisorctl`, there are also a lot of third party tools that use the API directly, so I don't think we can accept a solution which breaks all of them.

I agree on this.

But this issue exists for 8 years now - can there really nothing be done?

PR #1319, opened 2020-01-15
add simple bash completion script

bubak4:
- script is ready to be placed in '/etc/bash_completion.d/'
- based on https://gist.github.com/pfreixes/3187511 (+ some additional features)


mnaberez:
Thanks for this patch but I would rather not add this sort of thing to the core of Supervisor.  There is nothing else in Supervisor that depends on a particular shell and we also don't have a way to run this in our test suite to ensure it works and keeps working.

Issue #1327, opened 2020-02-09
`test_end_to_end.py:test_issue_1231c` fails on Python 3.6


mnaberez:
https://travis-ci.org/Supervisor/supervisor/jobs/622789487

This test seems to consistently fail on Travis now but it was working when #1235 was merged.  I restarted the build a few times.  I only observed the failure on 3.6 and only on that line.

Related #1235
cc @vsajip

vsajip:
Really weird. I stepped into `pexpect` to see what `supervisord` was sending. I haven't yet found out why, but the `supervisord` process is terminating prematurely after writing this out:

> Error: Another program is already listening on a port that one of our HTTP servers is configured to use.  Shut this program down first before starting supervisord.
> For help, use /home/vinay/projects/supervisor/supervisor/supervisord.py -h

It's not immediately clear what's going on. With Python 3.7, at the same point, we see

> Unlinking stale socket /tmp/issue-1231.sock
> 2020-02-10 20:43:51,341 INFO RPC interface 'supervisor' initialized
> 2020-02-10 20:43:51,341 CRIT Server 'unix_http_server' running without any HTTP authentication checking
> 2020-02-10 20:43:51,343 INFO supervisord started with pid 37469
> 2020-02-10 20:43:52,351 INFO spawned: 'hello' with pid 37472
> 2020-02-10 20:43:54,323 INFO success: hello entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)

vsajip:
I've done a `git bisect`, but that's inconclusive - it just identifies the bad commit as the one where I added `test_issue_1231c`, on 28 May 2019. So if it was working with Python 3.6 at that time, it doesn't any more. But the Python version I'm using predates that (22 Oct 2018), so I would guess that it's something else in the environment that's problematic (given that 3.7, etc. aren't failing).

vsajip:
OK, I've attempted a fix in #1329.

mnaberez:
@vsajip Thanks for looking into this!  I didn't run it locally yet but I looked at your patch #1329 and I agree with the changes, especially ensuring process shutdown in each test.

> I would guess that it's something else in the environment that's problematic (given that 3.7, etc. aren't failing).

[...]

PR #1329, opened 2020-02-11
First cut at a fix for #1327

vsajip:
The problem appears to be caused by three separate tests using the same config file, and hence the same socket for communication between `supervisorctl` and `supervisord`. The failure occurs because a socket is still in use from a previous test (child processes have been told to terminate by the previous test, but not died and cleaned up when the next test starts). It doesn't fail on other Python versions presumably because of timing differences.

The solution is to use different socket files for communication for each test.


mnaberez:
I cherry-picked 075b8a8 from above so master (f789a432109c204bb057b18599159d8eaaae198c) is now running that with [`pexpect==4.7.0`](https://github.com/Supervisor/supervisor/issues/1327#issuecomment-585315418).  I couldn't merge the commit that waits for the process to exit (but I want to) because [it hangs](https://travis-ci.org/Supervisor/supervisor/builds/648810399) on macOS for some reason.

mnaberez:
I wasn't able to solve the [hang](https://github.com/Supervisor/supervisor/pull/1329#issuecomment-586635140) on macOS because seems to occur on some versions of `pexpect` and not others.  I'm going to leave it for now instead of pinning to a certain version of `pexpect`.  I pulled in all the other changes from this patch and also made other commits to try and prevent cross-talk between the end-to-end tests.  Thanks!

PR #1330, opened 2020-02-12
redesign action buttons

dmytrokarpovych:
Change action button styling in a more modern way.
Replace background images to CSS styles.

Before:
image

After:
image


mnaberez:
Thanks!

Issue #1273, opened 2019-08-28
'ascii' codec can't decode byte 0x89 in position 170


Rabbit021:
**my exceptions** 
i am open **inet_http_server** for web interface，but i recieve an exception like this.

2019-08-28 11:32:42,060 ERRO :Server Error: <type 'exceptions.UnicodeDecodeError'>, 'ascii' codec can't decode byte 0x89 in position 170: ordinal not in range(128): file: /usr/lib/python2.7/site-packages/supervisor/http.py line: 380

**my enviment**
![image](https://user-images.githubusercontent.com/2327876/63824283-2893fb00-c989-11e9-9506-af5e497a8a08.png)
![image](https://user-images.githubusercontent.com/2327876/63824306-35b0ea00-c989-11e9-8b7a-669a5d30f7f9.png)
![image](https://user-images.githubusercontent.com/2327876/63824370-71e44a80-c989-11e9-8e7a-5c00d38b25a1.png)

marzdgzmn:
@Rabbit021 Any workaround you've currently implemented for this one?

mnaberez:
cc @vsajip

vsajip:
Please provide a minimal script (or set of scripts) and a `supervisord.conf` which allows replication of the problem.

pradeepcep:
I haven't looked at the surrounding code, but here's some additional info, if it helps. I tried to log the values of `self.ac_out_buffer` and `data` at `supervisor/http.py:380` to a file, to see what was causing the issue. While `self.ac_out_buffer` was empty, `data` had something similar to this:
```

HTTP/1.1 200 OK
Date: Tue, 10 Dec 2019 11:47:27 GMT
Last-Modified: Thu, 21 Nov 2019 12:53:08 GMT
Content-Length: 1702
Content-Type: image/png
Server: Medusa/1.12

�PNG

IHDR szzbKGD�������     pHYs
```

I was using the web interface from Google Chrome when this error was logged. So I guess it's trying to operate on binary data as if it were a string?

```
vagrant@ubuntu-xenial:/vagrant$ python -V
Python 2.7.12
```

jpralves:
It seems that it triggers the error when loading binary resources like:
```
/images/supervisor.gif
/images/rule.gif
/images/icon.png
```

jpralves:
It seems that the affected line (http.py:380) can be replaced by:
```
self.ac_out_buffer =  self.ac_out_buffer + str(data)
```
[...]

PR #1347, opened 2020-05-10
First cut at fix for #1273.

vsajip:
IMO the problem is caused by the response body being generated as Unicode rather than bytes. Here is the proposed fix. There are a couple of other cases of `as_string()` in `web.py` which could also potentially be switched over, as we should generally be always passing bytes around. But I haven't implemented those changes, as they're not part of the bug report.


mnaberez:
Thanks!

PR #1187, opened 2018-12-25
Freezing zombie channels into a list avoids RuntimeError

cclauss:
Fixes #1178 which says that a ___RuntimeError___ is raised when the dictionary size changes during iteration.  This change freezes dict.values() into a list before the loop begins so that the list will not change during iteration. 🎄

PR #1352, opened 2020-06-04
Close channel  error

JesseJ12345:
Fix error: dictionary changed size during iteration

### DIFF
--- a/supervisor/medusa/http_server.py
+++ b/supervisor/medusa/http_server.py
@@ -506,10 +506,13 @@ class http_channel (asynchat.async_chat):
 
     def kill_zombies (self):
         now = int (time.time())
-        for channel in list(asyncore.socket_map.values()):
-            if channel.__class__ == self.__class__:
-                if (now - channel.last_used) > channel.zombie_timeout:
-                    channel.close()
+        channels = list(asyncore.socket_map.values())
+        chan_len = len(channels)
+        if chan_len:
+            for i in range(chan_len):
+                if channels[i].__class__ == self.__class__:
+                    if (now - channel[i].last_used) > channel[i].zombie_timeout:
+                        channel[i].close()
 
     # --------------------------------------------------
     # send/recv overrides, good place for instrumentation.
--- a/supervisor/medusa/http_server.py
+++ b/supervisor/medusa/http_server.py
@@ -511,8 +511,8 @@ class http_channel (asynchat.async_chat):
         if chan_len:
             for i in range(chan_len):
                 if channels[i].__class__ == self.__class__:
-                    if (now - channel[i].last_used) > channel[i].zombie_timeout:
-                        channel[i].close()
+                    if (now - channels[i].last_used) > channels[i].zombie_timeout:
+                        channels[i].close()
 
     # --------------------------------------------------
     # send/recv overrides, good place for instrumentation.


mnaberez:
> Fix error: dictionary changed size during iteration

Please tell us how to reproduce this error.  What do we need to do to make this error occur?

supervisor/medusa/http_server.py
```diff
@@ -506,10 +506,13 @@ def maintenance (self):
 
     def kill_zombies (self):
         now = int (time.time())
-        for channel in list(asyncore.socket_map.values()):
-            if channel.__class__ == self.__class__:
-                if (now - channel.last_used) > channel.zombie_timeout:
-                    channel.close()
+        channels = list(asyncore.socket_map.values())
+        chan_len = len(channels)
```
mnaberez, line 510: > Fix dictionary changed size during iteration

This code does not have a dictionary.  I also don't see why this change would make a difference, since `list(asyncore.socket_map.values())` should not change.  Can you please explain?

JesseJ12345:
> > Fix error: dictionary changed size during iteration
> 
> Please tell us how to reproduce this error. What do we need to do to make this error occur?

When kill_zombies triggered and timeout, it will close the channel(delete channel from the fd dict(asyncore.socket_map), close socket), but at the same time,  we can not use 'for' to deal

mnaberez:
> When kill_zombies triggered and timeout, it will close the channel(delete channel from the fd dict(asyncore.socket_map), close socket), but at the same time, we can not use 'for' to deal

Could you please respond to the [comment above](https://github.com/Supervisor/supervisor/pull/1352#discussion_r435008790)?  I don't see how this error could occur or why this change would make a difference.

supervisor/medusa/http_server.py
JesseJ12345: oh, yes, you are right , my mistakes, it is the same. sorry about that.

since i met a bug :
```python
ERRO 0.0.0.0:uncaptured python exception, closing channel <supervisor.http.supervisor_af_inet_http_server listening 0.0.0.0:9001 at 0xb5fc5410> (<class 'RuntimeError'>:dictionary changed size during iteration 
...
http_server.py|kill_zombies|509])
```
Then i check the source code and ... . maybe this is another bug.
ps, my version is 4.0.0

THANKS FOR YOUR ANSWER

mnaberez: > ps, my version is 4.0.0

In Supervisor 4.0.0, the line is [the same](https://github.com/Supervisor/supervisor/blob/4.0.0/supervisor/medusa/http_server.py#L509).  It does not iterate a dictionary, so I don't see how it could produce the error `dictionary changed size during iteration`.

Perhaps you have a development version like `4.0.0.dev0` from before we merged [this fix](https://github.com/Supervisor/supervisor/pull/1187/files)(?).  Can you please check line 509 in `http_server.py` of your local copy?

JesseJ12345: YES, this is a error for my version
image

Ok, then this bug has solved in new version

Thanks a lot.

mnaberez: Great, thanks for following up!  I'll close this PR.

mnaberez:
Fixed in 4.0.0 (see https://github.com/Supervisor/supervisor/pull/1352#discussion_r435026744).

ghulamm777:
Great Allah always pray for you

PR #1355, opened 2020-06-28
Fix link when hosted under a subdirectory

brglng:
Fix the link when supervisor is hosted under a subdirectory behind nginx reverse proxy.


mnaberez:
Closing in favor of #593

Issue #1361, opened 2020-07-31
TypeError: can't concat str to bytes 


s4g:
This happens once in a couple of weeks on my always on system.

error: uncaptured python exception, closing channel <supervisor.http.deferring_http_channel connected '' at 0x7f8b801977f0 channel#: 37 requests:1> (<class 'TypeError'>:can't concat str to bytes [/.local/work/local/venv/vr-py382-base/lib/python3.8/site-packages/supervisor/supervisord.py|runforever|220] [/.local/work/local/venv/vr-py382-base/lib/python3.8/site-packages/supervisor/medusa/asyncore_25.py|handle_read_event|391] [/.local/work/local/venv/vr-py382-base/lib/python3.8/site-packages/supervisor/medusa/asynchat_25.py|handle_read|94])

Apparently related to Python 2 to 3 migration. Have to move back to Python 2.7.

Supervisor 4.2.0 (same on 4.1)

mnaberez:
Traceback reformatted for easier reading:

```
<supervisor.http.deferring_http_channel connected '' at 0x7f8b801977f0 channel#: 37 requests:1> 
(<class 'TypeError'>:can't concat str to bytes 
[/.local/work/local/venv/vr-py382-base/lib/python3.8/site-packages/supervisor/supervisord.py|runforever|220] 
[/.local/work/local/venv/vr-py382-base/lib/python3.8/site-packages/supervisor/medusa/asyncore_25.py|handle_read_event|391] 
[/.local/work/local/venv/vr-py382-base/lib/python3.8/site-packages/supervisor/medusa/asynchat_25.py|handle_read|94])
```

Possibly related to #1273 (not sure about that, cc @vsajip)

vsajip:
@s4g any chance you could try this simple change with Python 3.x and see if it resolves the issue?

s4g:
> @s4g any chance you could try this simple change with Python 3.x and see if it resolves the issue?

I patched the code. Will have to wait for at least a month to see if it repeats.

Thanks

mnaberez:
> @s4g any chance you could try this simple change with Python 3.x and see if it resolves the issue?

Thanks for the patch!  I went ahead and merged it.  Even if it does not fix this specific issue, it likely fixes a different one.  I intend to make a new release with it shortly.

mnaberez:
Fix released in [Supervisor 4.2.1](https://pypi.org/project/supervisor/4.2.1/).

[...]

PR #1362, opened 2020-08-06
Add possible fix for #1361: fix path which returns str rather than by…

vsajip:
…tes when a network error occurs.


mnaberez:
Thanks!

Issue #1346, opened 2020-05-03
SocketManagerTest.test_prepare_socket fails with AssertionError: 128 != 4096


sbraz:
Hi,
This was initially reported [here](https://bugs.gentoo.org/711538) but I wasn't able to reproduce it. Now I see the issue on one computer but not the others. It is present with Python 3.7 and 3.8.

It can be reproduced with `pytest -vv supervisor/tests/test_socket_manager.py::SocketManagerTest::test_prepare_socket` on the git master:

```
======================================= test session starts ========================================
platform linux -- Python 3.7.7, pytest-5.4.1, py-1.8.0, pluggy-0.13.1 -- /usr/bin/python3.7
cachedir: .pytest_cache
rootdir: /tmp/supervisor
collected 1 item                                                                                   

supervisor/tests/test_socket_manager.py::SocketManagerTest::test_prepare_socket FAILED       [100%]

============================================= FAILURES =============================================
______________________________ SocketManagerTest.test_prepare_socket _______________________________

self = <supervisor.tests.test_socket_manager.SocketManagerTest testMethod=test_prepare_socket>

    def test_prepare_socket(self):
        conf = DummySocketConfig(1)
        sock_manager = self._makeOne(conf)
        sock = sock_manager.get_socket()
        self.assertTrue(sock_manager.is_prepared())
        self.assertFalse(sock.bind_called)
        self.assertTrue(sock.listen_called)
>       self.assertEqual(sock.listen_backlog, socket.SOMAXCONN)
E       AssertionError: 128 != 4096

supervisor/tests/test_socket_manager.py:207: AssertionError
===================================== short test summary info ======================================
FAILED supervisor/tests/test_socket_manager.py::SocketManagerTest::test_prepare_socket - Assertio...
======================================== 1 failed in 0.12s =========================================
```

mnaberez:
This test [passes on CI](https://travis-ci.org/github/supervisor/supervisor/builds/681651598) on Python 3.4 - 3.8.  I ran it locally on a couple of systems and it passes there as well.  Are you able to give us any more information that might allow us to reproduce it?

Ma27:
@sbraz do you use glibc 2.31? I experienced the same issue there as well.

sbraz:
@mnaberez I'm sorry, it looks like I completely missed your message. I'm afraid I don't have any more info other than the fact that I'm running Gentoo.
@Ma27 yes I am using glib 2.31 too, maybe this is related. What distro do you use?

[...]

PR #1363, opened 2020-08-06
Reflect new glibc 2.31 default value for SOMAXCONN on tests (closes: #1346)

samueloph:
closes: #1346

### DIFF
--- a/supervisor/tests/base.py
+++ b/supervisor/tests/base.py
@@ -358,7 +358,7 @@ class DummySocketConfig:
         return not self.__eq__(other)
 
     def get_backlog(self):
-        return 128
+        return 4096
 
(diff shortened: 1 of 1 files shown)


mnaberez:
Will this change now cause it to fail on older glibc instead?

samueloph:
@mnaberez Yes, it will

samueloph:
hmm, would it be good solution to make the test check to either 128 or 4096?

mnaberez:
Thanks for your help and this patch.  I have just committed 48fd49eb857c2eaf5dcb45ff100d23ca7d41b4e9 which I believe will fix it.  Please test this if you are able and leave a comment on #1346 if it still has issues.

Issue #1373, opened 2020-09-24
Add option to exit supervisord if a program exits or can't restart


frankjas:
If supervisord is managing processes in a container and is set as the entrypoint in no-daemon mode then under certain circumstances it is desirable to have supervisord exit if a certain process exits.  For example, consider a container offering an HTTP api implemented in Python.  The container may have nginx, uwsgi, and a python daemon performing the app logic.

nginx and uwsgi should be restarted automatically, but if for some reason they can't restart then the container should exit since it no longer can serve HTTP requests.

The proposal is to add a config option for each process section which when true would trigger  a supervisord exit.

mnaberez:
Duplicate of #712

Issue #712, opened 2016-01-18
shutdown supervisord once one of the programs is killed


EslamElHusseiny:
Hi,
I've two programs managed by supervisord.
Is there anyway to shutdown supervisord once of those programs is killed ?

tarrychk:
You can achive this with an event listener:

supervisord-watchdog:
```
#!/usr/bin/python
import sys
import os
import logging
import subprocess
import time

from supervisor.childutils import listener

def main(args):
    logging.basicConfig(stream=sys.stderr, level=logging.DEBUG, format='%(asctime)s %(levelname)s %(filename)s: %(message)s')
    logger = logging.getLogger("supervisord-watchdog")
    debug_mode = True if 'DEBUG' in os.environ else False

    while True:
        logger.info("Listening for events...")
        headers, body = listener.wait(sys.stdin, sys.stdout)
        body = dict([pair.split(":") for pair in body.split(" ")])

        logger.debug("Headers: %r", repr(headers))
        logger.debug("Body: %r", repr(body))
        logger.debug("Args: %r", repr(args))

        if debug_mode: continue

        try:
            if headers["eventname"] == "PROCESS_STATE_FATAL":
                logger.info("Process entered FATAL state...")
                if not args or body["processname"] in args:
                    logger.error("Killing off supervisord instance ...")
                    res = subprocess.call(["/bin/kill", "-15", "1"], stdout=sys.stderr)
                    logger.info("Sent TERM signal to init process")
                    time.sleep( 5 )
                    logger.critical("Why am I still alive? Send KILL to all processes...")
                    res = subprocess.call(["/bin/kill", "-9", "-1"], stdout=sys.stderr)
        except Exception as e:
            logger.critical("Unexpected Exception: %s", str(e))
            listener.fail(sys.stdout)
            exit(1)
        else:
            listener.ok(sys.stdout)

if __name__ == '__main__':
    main(sys.argv[1:])
```

Then add this config snippet:
```
[eventlistener:supervisord-watchdog]
command=/path/to/supervisord-watchdog
events=PROCESS_STATE_FATAL
```

tarrychk:
That eventlistener snippet above was written for running supervisord in a docker container where it would have PID 1. You'll have to replace the kill commands with something suitable for your use case, like ```"pkill", "-15", "supervisord"```.

mnaberez:
[...]

PR #1374, opened 2020-09-24
Add option to exit supervisord on exit (or bad restart) of the config…

frankjas:
…ured processes

Instead of writing an event listener, the following new config is an alternative and is perhaps simpler to setup.


mnaberez:
Please see comments in #1373 and #712.  As stated there are no plans to add features to cause supervisord to exit itself based on various conditions.

Issue #1375, opened 2020-10-09
stdout_syslog depends on stdout_logfile


scottkof:
**Expected behavior:** `stdout_syslog` and `stdout_logfile` are independent.
Given the following configs
```
  stdout_logfile = NONE
  stdout_syslog = true
```
child process logs should be logged to syslog.
<hr>

**Actual behavior:** `stdout_syslog` depends on `stdout_logfile`.
Given the following configs
```
  stdout_logfile = NONE
  stdout_syslog = true
```
child process logs are _not_ logged to syslog.

However, given the following configs
```
  stdout_logfile = AUTO
  stdout_syslog = true
```
child process logs are written to syslog (and logfile).

scottkof:
[PR #1376](https://github.com/Supervisor/supervisor/pull/1376) is a fix for this issue.

mnaberez:
Fixed in #1376 (committed as ab1e07cc6d502c452f00c1e2d88989e91e75f294).

PR #1376, opened 2020-10-09
Decouple stdout_syslog and stdout_logfile (see Issue #1375)

scottkof:
This is a fix for [Issue #1375](https://github.com/Supervisor/supervisor/issues/1375)

### DIFF
--- a/supervisor/dispatchers.py
+++ b/supervisor/dispatchers.py
@@ -129,6 +129,8 @@ class POutputDispatcher(PDispatcher):
         """
 
         logfile = getattr(config, '%s_logfile' % channel)
+        if logfile == '':
+            return
 
         maxbytes = getattr(config, '%s_logfile_maxbytes' % channel)
         backups = getattr(config, '%s_logfile_backups' % channel)
--- a/supervisor/dispatchers.py
+++ b/supervisor/dispatchers.py
@@ -129,8 +129,6 @@ class POutputDispatcher(PDispatcher):
         """
 
         logfile = getattr(config, '%s_logfile' % channel)
-        if not logfile:
-            return
 
         maxbytes = getattr(config, '%s_logfile_maxbytes' % channel)
         backups = getattr(config, '%s_logfile_backups' % channel)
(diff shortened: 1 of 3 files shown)


supervisor/dispatchers.py
```diff
@@ -139,17 +137,23 @@ def _setup_logging(self, config, channel):
...
                 "Use %s_syslog instead." % channel, DeprecationWarning)
             fmt = ' '.join((config.name, fmt))
-        self.mainlog = loggers.handle_file(
-            config.options.getLogger(),
-            filename=logfile,
-            fmt=fmt,
-            rotating=not not maxbytes, # optimization
-            maxbytes=maxbytes,
-            backups=backups)
+        if logfile is not None:
```
mnaberez, line 140: This replaces the original comparison `if not logfile` which would evaluate to `False` on both `None`and an empty string.  This data originates from parsing the config file, so it an empty string may actually be possible (unsure).  Was this change intentional, and if so are you certain there is no code path where this may receive an empty string?

supervisor/loggers.py
```diff
@@ -406,6 +406,8 @@ def handle_syslog(logger, fmt):
     handler.setLevel(logger.level)
     logger.addHandler(handler)
 
+    return logger
+
```
mnaberez, line 410: It is weird that some of these `handle_` functions return a `logger` and some don't.  Why was this change made?

supervisor/tests/test_dispatchers.py
```diff
@@ -512,6 +512,76 @@ def test_syslog_logfile_deprecated(self):
             self._makeOne(process)
             self.assertEqual(len(w), 1)
 
+    def test_no_logfile_and_no_syslog(self):
+        from supervisor.datatypes import boolean, logfile_name
```
mnaberez, line 516: No other existing tests in this file import things from `supervisor.datatypes` although they do similar configuration setup.

supervisor/tests/test_dispatchers.py
```diff
@@ -512,6 +512,76 @@ def test_syslog_logfile_deprecated(self):
...
+        from supervisor.options import ProcessConfig, ServerOptions
+        options = ServerOptions() # need real options to get a real logger
+        options.loglevel = 20 # INFO, see loggers.py LevelsByName
```
mnaberez, line 519: Existing tests in this file do things like `options.loglevel = loggers.LevelsByName.TRAC`.  Why is this different?

mnaberez:
Thanks for the excellent original report and this nice patch.  Some minor questions above.

supervisor/loggers.py
scottkof: I made this change because I needed a logger at dispatchers.py:154. Maybe loggers are only returned from the handle_ functions when they're needed?

supervisor/tests/test_dispatchers.py
scottkof: Other tests don't read special LOGFILE_ datatypes (datatypes.py:91-92) into their configs. I used the datatypes functions so that the test config input (e.g. lines 521-22) is exactly what would be read from a real config file. I believe this makes the tests more readable and more robust.

supervisor/tests/test_dispatchers.py
scottkof: This was an oversight. I'll change it.

supervisor/dispatchers.py
scottkof: The code now specially handles `logfile` being an empty string, in which case the behavior is the same as always.

mnaberez:
Committed in ab1e07cc6d502c452f00c1e2d88989e91e75f294.  Thanks!

PR #1402, opened 2021-02-02
Ignore current uid when dropping privileges

aeolusp:
**Problem**:
If starting the process as a different user from supervisor then it is mandatory to run supervisor as root. This is because the current code checks for supervisor uid=0 i.e. root. However, on Linux systems, it is sufficient to have cap_setuid and cap_setgid to set effective user and group ids. 

**Solution**:
Instead of checking for root UID (i.e. 0), it is sufficient to check the exception from setuid. 

**Scenario**:
From supervisor start a process that runs with a different user account.
If running with privileged user i.e. root, -> success
If running supervisor as an unprivileged user,
a) with no capabilities -> fails
b) with capability cap_setuid alone -> fails
c) with capability cap_setgid alone -> fails
d) with capabilities cap_setuid and cap_setgid both -> succeeds

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1372,6 +1372,11 @@ class ServerOptions(Options):
             # has "user=foo" (same user) in it.
             return
 
+        try:
+            os.setuid(uid)
+        except:
+            return 'Could not set user id of effective user'
+
         gid = pwrec[3]
         if hasattr(os, 'setgroups'):
             user = pwrec[0]
@@ -1392,11 +1397,6 @@ class ServerOptions(Options):
             os.setgid(gid)
         except OSError:
             return 'Could not set group id of effective user'
-
-        try:
-            os.setuid(uid)
-        except:
-            return 'Could not set user id of effective user'
 
     def set_uid_or_exit(self):
         """Set the uid of the supervisord process.  Called during supervisord
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1396,7 +1396,7 @@ class ServerOptions(Options):
         try:
             os.setuid(uid)
         except:
-            return 'Could not set effective user id'
+            return 'Could not set user id of effective user'
 
     def set_uid_or_exit(self):
         """Set the uid of the supervisord process.  Called during supervisord
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1372,11 +1372,6 @@ class ServerOptions(Options):
             # has "user=foo" (same user) in it.
             return
 
-        try:
-            os.setuid(uid)
-        except:
-            return 'Could not set user id of effective user'
-
         gid = pwrec[3]
         if hasattr(os, 'setgroups'):
             user = pwrec[0]
@@ -1397,6 +1392,11 @@ class ServerOptions(Options):
             os.setgid(gid)
         except OSError:
             return 'Could not set group id of effective user'
+
+        try:
+            os.setuid(uid)
+        except:
+            return 'Could not set user id of effective user'
 
     def set_uid_or_exit(self):
         """Set the uid of the supervisord process.  Called during supervisord
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1372,9 +1372,6 @@ class ServerOptions(Options):
             # has "user=foo" (same user) in it.
             return
 
-        if current_uid != 0:
-            return "Can't drop privilege as nonroot user"
-
         gid = pwrec[3]
         if hasattr(os, 'setgroups'):
             user = pwrec[0]
@@ -1395,7 +1392,11 @@ class ServerOptions(Options):
             os.setgid(gid)
         except OSError:
             return 'Could not set group id of effective user'
-        os.setuid(uid)
+
+        try:
+            os.setuid(uid)
+        except:
+            return 'Could not set effective user id'
 
     def set_uid_or_exit(self):
         """Set the uid of the supervisord process.  Called during supervisord
--- a/supervisor/tests/test_options.py
+++ b/supervisor/tests/test_options.py
@@ -3253,7 +3253,7 @@ class ServerOptionsTests(unittest.TestCase):
     def test_drop_privileges_nonroot_different_user(self):
         instance = self._makeOne()
         msg = instance.drop_privileges(42)
-        self.assertEqual(msg, "Can't drop privilege as nonroot user")
+        self.assertEqual(msg, "Could not set user id of effective user")
 
     def test_daemonize_notifies_poller_before_and_after_fork(self):
         instance = self._makeOne()
--- a/supervisor/tests/test_options.py
+++ b/supervisor/tests/test_options.py
@@ -3253,7 +3253,7 @@ class ServerOptionsTests(unittest.TestCase):
     def test_drop_privileges_nonroot_different_user(self):
         instance = self._makeOne()
         msg = instance.drop_privileges(42)
-        self.assertEqual(msg, "Could not set user id of effective user")
+        self.assertEqual(msg, "Could not set group id of effective user")
 
     def test_daemonize_notifies_poller_before_and_after_fork(self):
         instance = self._makeOne()


mnaberez:
[Build failures](https://travis-ci.org/github/Supervisor/supervisor/builds/757288051):
> ```
> =================================== FAILURES ===================================
> ________ ServerOptionsTests.test_drop_privileges_nonroot_different_user ________
> self = <supervisor.tests.test_options.ServerOptionsTests testMethod=test_drop_privileges_nonroot_different_user>
>     @patch('pwd.getpwuid', Mock(return_value=["foo", None, 55, 34]))
>     @patch('os.getuid', Mock(return_value=12))
>     def test_drop_privileges_nonroot_different_user(self):
>         instance = self._makeOne()
>         msg = instance.drop_privileges(42)
> >       self.assertEqual(msg, "Can't drop privilege as nonroot user")
> E       AssertionError: 'Could not set groups of effective user' != "Can't drop privilege as nonroot user"
> E       - Could not set groups of effective user
> E       + Can't drop privilege as nonroot user
> supervisor/tests/test_options.py:3256: AssertionError
> ==================== 1 failed, 1383 passed in 56.89 seconds ====================
> ERROR: InvocationError for command /home/travis/build/Supervisor/supervisor/.tox/py34/bin/py.test (exited with code 1)
> ```

mnaberez:
Additional information was provided in #1405:

> This is related to the fix - #1405 
> # Setup:
> 1) Running `supervisord` as a non-root user for eg. `user1`
> 2) Launch another process from `supervisor` with another non-root user for eg. `user2`
> 3) the required capabilities `cap_setuid` and `cap_setgid` assigned for `python` (`supervisord` is script and not executable)
> 
> Start `supervisord`
> ```
> # su - user1
> [user1@TESTSYS /home/user1]
> # /usr/bin/python3 /usr/bin/supervisord -c /home/user1/superv.conf
> .
> .
> supervisor: couldn't setuid to 620: Can't drop privilege as nonroot user
> ```
> Although required capabilities set it is not possible to launch process with different non-root user is `supervisord` is running as a non-root user.

supervisor/options.py
```diff
@@ -1395,7 +1397,6 @@ def drop_privileges(self, user):
             os.setgid(gid)
         except OSError:
             return 'Could not set group id of effective user'
-        os.setuid(uid)
```
mnaberez, line 1398: Why specifically are you trying to move `os.setuid()` before setting `os.setgroups()` and `os.setgid()`?

aeolusp: with previous code this the place where the code used to break if there is a problem with setuid. So, I moved setuid before so that if there is a problem with setuid it breaks at the same place. In fact, the order setuid or setgid should not matter.

mnaberez: Sorry, I don't understand this explanation.  If I am reading correctly, you are saying that you moved it because it was broken in the place that it was, but then you also said you think the order does not matter.

aeolusp:
1) build failed because test case failed. Test case failed because the
returned error string changed.
2) to did failure the treat case updated
3) moved setuid before because to detect failure at earliest please. In
other words fittings performs other tasks only to fail at end, not a good
idea!

aeolusp:
Auto correction screwed up sentence.. Correcting below.
1) build failed because test case failed. Test case failed because the
returned error string changed.
2) to fix failure the test case updated
3) moved setuid before because to detect failure at earliest please. In
other words of kept at end means, function performs other tasks only to
fail at end, not a good idea!

supervisor/options.py
mnaberez: https://github.com/Supervisor/supervisor/pull/1402#issuecomment-772720968:
> 3) moved setuid before because to detect failure at earliest please. In other words of kept at end means, function performs other tasks only to fail at end, not a good idea!

I don't understand how it could work if `os.setuid()` is called before `os.setgroups()` and `os.setgid()` as you propose here.  I think the order is important and that the order needs to be as it was originally.

Before any code changes:

```text
$ id unprivileged 
uid=1001(unprivileged) gid=1001(unprivileged) groups=1001(unprivileged)

$ id 
uid=0(root) gid=0(root) groups=0(root)

$ whoami 
root 

$ cat supervisord.conf 
[supervisord]
nodaemon=true 
loglevel=debug 

[program:id]
user = unprivileged
command = /usr/bin/id 
startsecs = 0 

$ supervisord -c supervisord.conf 
2021-02-03 11:02:35,141 CRIT Supervisor is running as root.  Privileges were not dropped because no user is specified in the config file.  If you intend to run as root, you can set user=root in the config file to avoid this message.
2021-02-03 11:02:35,145 INFO supervisord started with pid 2663
2021-02-03 11:02:36,148 INFO spawned: 'id' with pid 2666
2021-02-03 11:02:36,158 DEBG 'id' stdout output:
uid=1001(unprivileged) gid=1001(unprivileged) groups=1001(unprivileged)
```

Making no other changes, other than to move `os.setuid()` before `os.setgroups()` and `os.setgid()` as you have done in this pull request:

```diff
diff --git a/supervisor/options.py b/supervisor/options.py
index 273d8a5..c83fe44 100644
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1375,6 +1375,8 @@ class ServerOptions(Options):
         if current_uid != 0:
             return "Can't drop privilege as nonroot user"
 
+        os.setuid(uid)
+
         gid = pwrec[3]
         if hasattr(os, 'setgroups'):
             user = pwrec[0]
@@ -1395,7 +1397,6 @@ class ServerOptions(Options):
             os.setgid(gid)
         except OSError:
             return 'Could not set group id of effective user'
-        os.setuid(uid)
 
     def set_uid_or_exit(self):
         """Set the uid of the supervisord process.  Called during supervisord
```

Results in the call to `os.setgroups()` failing:

```
$ whoami 
root 

$ supervisord -c supervisord.conf 
2021-02-03 11:05:07,287 CRIT Supervisor is running as root.  Privileges were not dropped because no user is specified in the config file.  If you intend to run as root, you can set user=root in the config file to avoid this message.
2021-02-03 11:05:07,290 INFO supervisord started with pid 2706
2021-02-03 11:05:08,293 INFO spawned: 'id' with pid 2709
2021-02-03 11:05:08,300 DEBG 'id' stderr output:
supervisor: couldn't setuid to 1001: Could not set groups of effective user

2021-02-03 11:05:08,301 INFO success: id entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2021-02-03 11:05:08,301 DEBG 'id' stderr output:
supervisor: child process was not spawned

2021-02-03 11:05:08,303 DEBG fd 8 closed, stopped monitoring <POutputDispatcher at 140498999981840 for <Subprocess at 140499000370624 with name id in state RUNNING> (stderr)>
2021-02-03 11:05:08,303 DEBG fd 6 closed, stopped monitoring <POutputDispatcher at 140499007533408 for <Subprocess at 140499000370624 with name id in state RUNNING> (stdout)>
2021-02-03 11:05:09,306 INFO exited: id (exit status 127; not expected)
2021-02-03 11:05:09,306 DEBG received SIGCLD indicating a child quit
```

aeolusp: yes, you got it right. Apologies I overlooked that previous check. the setuid must be called only after fetching all needed details of the current user. I put the setuid at the end as before.

supervisor/options.py
```diff
@@ -1395,7 +1392,11 @@ def drop_privileges(self, user):
...
+
+        try:
+            os.setuid(uid)
+        except:
```
mnaberez, line 1398: This should be `except OSError` like the others.  

You added this `try`/`except` around `os.setuid()`.  Previously, there was none and this function would throw an exception from `os.setuid()`.  After this change, it will no longer do this.  Have you thoroughly checked that every possible code path that calls this function (`drop_privileges`) will not change behavior as a result of this change?

This change seems more consistent with the rest of the function, but it can't be made without carefully considering its impact on the rest of the program.

mnaberez:
Closing due to inactivity.

PR #1403, opened 2021-02-03
Improve the rationale of why distribution packages lag behind

Lin-Buo-Ren:
Signed-off-by: 林博仁(Buo-ren, Lin) <Buo.Ren.Lin@gmail.com>

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -85,7 +85,8 @@ the system boots.
 
 .. note::
 
-    Distribution packages of Supervisor can lag considerably behind the
+    Due to the software release policy set by each of the Linux distributions,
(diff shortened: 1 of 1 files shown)


mnaberez:
I think it is enough to say the main point, which is that the distribution packages can lag behind.  If the reader needs to understand why a particular distribution package happens to lag behind, they should seek the answer with that distribution.

Lin-Buo-Ren:
> If the reader needs to understand why a particular distribution package happens to lag behind, 
> they should seek the answer with that distribution.

Provide a simple answer will eliminate the need for such action, which can let the reader not distracted by non-related details, though.

PR #1404, opened 2021-02-03
Document BACKOFF and its delay mechanism

StarLightPL:
This is in response to github issues raised about configuring backoff
delay. This change just highlights that there is an increasing delay
between retries because it wasn't immediately obvious that it is
happening. Also the connection to startsecs / startretries was unclear.


mnaberez:
Thanks!

PR #1349, opened 2020-05-21
Append timestamps to logs

axil:
This is the second advent of PR #333

Issue #553, opened 2015-01-08
Add ability to specify log format


mminer:
It would be helpful to have control over the format of the log output. For inspiration, other tools that allow you to customize the log format include Celery ([CELERYD_LOG_FORMAT](http://celery.readthedocs.org/en/latest/configuration.html#celeryd-log-format)) and Gunicorn ([access_log_format](http://docs.gunicorn.org/en/19.1.1/settings.html#access-log-format)).

SebastianSzturo:
:+1: 
I would like to have heroku-style application name and output in one line.

avli:
+1

silvexis:
+1

LouisKottmann:
Any reason why this was closed? I'd like to remove the timestamps because syslog-ng does it already for example (and my log aggregator can understand syslog-ng ones)

+1

pearofducks:
@LouisKottmann - it looks like this is still open for Supervisor, and closed in the dokku-supervisord project (which just referenced this issue)

LouisKottmann:
Ah you're right, my bad

lciti:
+1
circusd, a similar project, also [has it](https://circus.readthedocs.io/en/latest/for-ops/configuration/): stdout_stream.time_format = %Y-%m-%d %H:%M:%S

Fire-Brand:
+1

nicbou:
We really need this. It's currently impossible to use supervisord with some logging tools because the format is completely random.

boblozano:
For us it's not impossible, just awkward. Of course because it's logging, the reminders are constant.

samirop:
Still no news regarding this?

Kimamisa:
+1

harpcio:
+1

isavl:
+1

isavl:
+1

dingman:
+1

Narcissist1:
+1

sconnor-USM:
I was able to get this working by using a shell script in the following manner (definitely would be nice to not have to do this workaround but it worked)

`
#!/bin/bash
WORKER_NAME=$1
bash -c 'php -f /PATH_TO_WORKER/"$WORKER_NAME".php | ts '"'"'[%d-%b-%Y %H:%M:%S]'"'"
`
and then the supervisor config command can look as such

`
[program:worker]
command=/PATH_TO_SHELL/run-worker.sh "Worker_NAME"
`
[...]

PR #1407, opened 2021-02-08
Option to prepend timestamp to stdout and stderr log file with config. 

the-c0d3r:
This is to revive the PR #333 , but this time, I have added proper unit testing for the output dispatcher. 

I retained all the original changes from the PR #333 and merged to current master, and added 2 new unit test functions and 4 new tests to the existing test function. 

Let me know if the unit testing is insufficient, or is there anything else to change. 

Related Issue: #553 Add ability to specify log format (6 years old issue)
Related PR: #1349 (closed due to lack of unit tests)
Credits to @nirmalc for the original code change.


mnaberez:
I was asked to review and merge this, but I don't really understand this subsystem enough to do that.  However, I did test it anyway, and it seems to break `supervisorctl fg`mode.  I'll leave this open until one of the other committers who understands this subsystem can work on it, or perhaps a future time when I understand it.

the-c0d3r:
Hi @mnaberez, may I ask what is the status with this PR?

Could you please tell me how it breaks `supervisorctl fg`? 
Based on my brief testing on the fg mode, it doesn't seem to have an issue.

mnaberez:
> Hi @mnaberez, may I ask what is the status with this PR?

https://github.com/Supervisor/supervisor/pull/1407#issuecomment-776080464

> Could you please tell me how it breaks `supervisorctl fg`?
> Based on my brief testing on the fg mode, it doesn't seem to have an issue.

The timestamps are mixed into the output, making fg mode unusable.

FazlanFarook:
waiting

the-c0d3r:
Based on my testing with simple bash script, I only get the timestamp if I set the `stdout_prepend_timestamp` to true. And when the timestamp is prepended, it also behaves as programmed.

I do not see issue with this, maybe you can elaborate more on this?

mnaberez:
> I do not see an issue with this, maybe you can elaborate more on this?

I don't think most users would expect changing a logging would cause unexpected output to be injected into fg mode.  It would make using a lot of software under fg mode unusable.  My opinion is that fg mode should probably work correctly regardless of which logging options are set.

ZPascal:
Hi community, 

thanks for the PR @the-c0d3r. I think it's a helpful feature, and we should solve the problem. Could you please explain how do you test the `fg` mode? Maybe we can solve the issue together and someone can merge the PR at the end of the process.

the-c0d3r:
Hi @ZPascal, 
I don't think that `fg` thing can be solved. I traced the `supervisorctl`'s fg mode, it calls HTTP api that is hosted by the `supervisord`, which will actually open the log file of the process and send it to the stdout from `supervisorctl`. So the only way to make the prepended timestamp *not* affect the `fg` mode is to change `supervisord`'s HTTP api to strip the timestamp when reading the log file, which I don't think is the way to go.

If you are looking for how to test the `fg` mode, here's how.
`config.conf` file content
```
[unix_http_server]
file=/tmp/supervisor.sock   ; the path to the socket file

[supervisord]
logfile=/tmp/supervisord.log ; main log file; default $CWD/supervisord.log
logfile_maxbytes=50MB        ; max main logfile bytes b4 rotation; default 50MB
logfile_backups=10           ; # of main logfile backups; 0 means none, default 10
loglevel=info                ; log level; default info; others: debug,warn,trace
pidfile=/tmp/supervisord.pid ; supervisord pidfile; default supervisord.pid
nodaemon=false               ; start in foreground if true; default false
silent=false                 ; no logs to stdout if true; default false
minfds=1024                  ; min. avail startup file descriptors; default 1024
minprocs=200                 ; min. avail process descriptors;default 200

[rpcinterface:supervisor]

[inet_http_server]
port = 127.0.0.1:9001

[supervisorctl]
serverurl=unix:///tmp/supervisor.sock ; use a unix:// URL  for a unix socket

[program:bash]
command=/tmp/test.sh
stdout_prepend_timestamp=true
```

`/tmp/test.sh` content
```sh
#!/bin/bash
while true; do
    echo "This is from test.sh $(date)";
    sleep 1;
done;
```

Run it as follows
```
supervisord -c config.conf
supervisorctl
> fg bash
```

---

As for the status of this PR, somewhere on the #553 , it was mentioned that this PR will not close it, as it only prepend timestamp, but the issue #553 is about specifying the log format. I will find some time to work on it and push the change for a customised timestamp format soon.

mnaberez:
> This is all I can do for this PR for now, as I have no idea how the fg mode is supposed to be fixed or whether it requires any fixes.

As I indicated [above](https://github.com/Supervisor/supervisor/pull/1407#issuecomment-803434908), I don't think users will except unexpected output being injected into `fg` mode, so from my perspective, it requires fixes.  I don't think we can add new features that break existing features.  It would be nice to add the feature but it would have to work with the rest of the system.

People use `fg`mode often, e.g. just recently in #1494 a question about `fg` mode was posted.  That screenshot in that issue shows a debugging session in `fg` mode, which I think would probably not be usable if this PR was merged and the user had the timestamp enabled.  I think the implementation of `fg` mode would have to change if something like this PR was to be merged.  Perhaps instead of tailing the logfile from disk (which would have timestamps mixed in), the daemon could keep a small circular buffer or use a temporary file for `fg` mode.  

Another thing that would have to be checked if is a patch like this would also interfere with the [events](http://supervisord.org/events.html) system.  There are events emitted for process log output and these must not have unexpected output mixed it.

I was reading https://github.com/Supervisor/supervisor/issues/553#issuecomment-1067752501 and I believe the problem described in that comment may apply here as well.  Depending on how/when the dispatcher flushes, the timestamp may not get added correctly.

ZPascal:
@the-c0d3r Thank you for the detailed explanation.

SergeySetti:
Ok, when?

PR #1410, opened 2021-02-20
Add two configuration properties to the getAllConfigInfo method

yellmean:
Add two configuration properties to the getAllConfigInfo method

### DIFF
--- a/supervisor/rpcinterface.py
+++ b/supervisor/rpcinterface.py
@@ -563,6 +563,7 @@ class SupervisorNamespaceRPCInterface:
             inuse = gconfig.name in self.supervisord.process_groups
             for pconfig in gconfig.process_configs:
                 d = {'autostart': pconfig.autostart,
+                     'directory': pconfig.directory,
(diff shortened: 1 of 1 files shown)


mnaberez:
> Add two configuration properties to the getAllConfigInfo method

Why are you requesting this change?

yellmean:
> > Add two configuration properties to the getAllConfigInfo method
> 
> Why are you requesting this change?

In order to facilitate the detection and management of the program, this time I just want to use these two properties, more importantly, you can more fully expose the relevant configuration information to the interface without the user to do more work.

mnaberez (changes requested):
Needs tests

mnaberez:
> In order to facilitate the detection and management of the program, this time I just want to use these two properties, more importantly, you can more fully expose the relevant configuration information to the interface without the user to do more work.

I am having difficulty following this.  Maybe you could explain the actual task you tried to do that required you to add these?  Thank you.

PR #1413, opened 2021-02-27
Stop running tests when building the documentation

sbraz:
Hi,
I got slightly confused when I noticed that building the documentation triggered a test run.

This reverts part of 674109eadc61943c593044d1375a0e8f159ae2ed which added the call to `setup.py test`, let me know if I missed something important but I don't see any reason to keep this weird behaviour.


mnaberez:
> I don't see any reason to keep this weird behaviour.

I don't either.  Merged, thanks!

PR #1415, opened 2021-03-03
kwargs not supported by pythons xmlrpc.client

wackazong:
It seems that kwargs are not supported by the xmlrpc library:

https://github.com/python/cpython/blob/b04f1cb9df7ad93366ef0ef7d8088effc576c5ae/Lib/xmlrpc/client.py#L1115

Too bad, because then you cannot use the `wait` named parameter. I just found that out the hard way, would be useful to quickly state thin in the docs I think.


wackazong:
Sorry, misunderstood docs. It's not a keyword argument, just a second Param.

PR #1426, opened 2021-04-20
Adding plaintext and JSON logging formatters

Alsheh:


ion-f3:
Hi. I'm quite interested in this PR and willing to contribute if needed. What are the next steps here? Looks like it needs some tests for the new functionality?

Alsheh:
hey @ion-f3 - thanks for expressing interest in this PR and your contribution is welcome! Next steps:
* Adding tests for new functionality.
* Python 2 backward compatibility to pass CI tests (only if this is a hard requirement by Supervisor).
* Investigate an issue where, sometimes, multiple lines are combined into one log message when the `JSON` formatter is used (e.g. `{"message": "first line output\nsecond line output"}`). The logger just emits whatever is given to it so my guess is that the dispatcher is combining these multiline messages; the issue is not obvious until the JSON formatter is turned on and the `\n` the character is included in between the combined log messages in the message field.

wvdschel-f3:
@mnaberez @mcdonc would you guys be open to merging this into the main branch at some point?

PR #1429, opened 2021-04-24
Add a new comments= option to [program:x] section

bbg221:
when I add chinese words to config file and show to the web page,  it always lost something in html-string make webpage shows strange in chrome.
after debug,  I found that num_sent can bigger than obs,  as num_sent is 4206 but obs is 4096. it means  4096 bytes send in the first time, but in the second time, it send from the  position 4207. that's the reason why something lost in the webpage.


mnaberez:
Can you please help us reproduce the problem so we can test this patch?

- Can you please give a sample config file that causes the problem?
- What webpage do we have to load, and what do we have to click on, to see the problem?
- What version of Python are you running?

Thanks for your help in solving this issue.

bbg221:
> Can you please help us reproduce the problem so we can test this patch?
> 
> * Can you please give a sample config file that causes the problem?
> * What webpage do we have to load, and what do we have to click on, to see the problem?
> * What version of Python are you running?
> 
> Thanks for your help in solving this issue.

Hello, the issue is coming after my changes on the project but not the original project. I forked this project and  commit the changes on it.  I just want to add chinese comments of each process in the configure file  and show it on the webpage(status.html). meanwhile I  commit the configure file and the issue picture and my debugging picture.
mayby utf8 words in the html will lead this issue. 
I use python 2.7.5  which is the detault version in centos7.
[https://github.com/bbg221/supervisor](https://github.com/bbg221/supervisor)

mnaberez:
Note for maintainers:
 - Adds a new `comments=` option to the `[program:x]` section of the config file.  The value of this new option is displayed on the web UI.  Also changes all text on the web UI from English to Chinese (e.g. button titles).
 
 - Includes changes that alter unicode behavior throughout `supervisord` (see `reload(sys)` in several places and medusa patches).  The `reload(sys)` in particular looks perhaps not correct, as `supervisord` handles unicode text elsewhere without such things.
 
 - Submitter indicates [above](https://github.com/Supervisor/supervisor/pull/1429#issuecomment-826279237) that the unicode-related changes are needed on Python 2.7.5 and were not needed before `comments=` was added.

bbg221:
I do not want to commit the furture "add comments" to master as many thing is not do very well now,  like code not runs well with python3, or user do not like the comments column in the web page as they familar with English and so on. I just commit it for reproduce the issue(my first commit). This is my first time to pull requests in github, all the submits are related to this pull requests is out of my mind.

PR #1446, opened 2021-06-25
feature: add support for runningregex to set RUNNING state accordingly

### LINKED ISSUE #1445: how to start dependent processes only when another child process is ready?
Hi,

I am facing some issues with setting the RUNNING state correctly. We need to delay state transition from starting to running. For example if a process needs 10 seconds to come up, the running should only be set when the process is actually running.

Is it possible to achieve this without the startsecs configuration, but only set it when the child process is actually ready?
Would appreciate any ideas on how to get this working or if there is a known workaround at the moment.
 
I am using supervisor 4.2.2. and I also created a suggestion for the implementation here: https://github.com/Supervisor/supervisor/pull/1446

Appreciate your help!

michaelhammann:
The runningregex option can be used to only set the RUNNING state of a process, if the child process is considered to be running. This is done by looking into the stdout and check whether an expected output has been recognized. If so, the process will set to running state. This requires the child process to create the corresponding output, otherwise the RUNNING state is never reached.

Closes #1445 

This option can be used by adding runningregex to the config of a program.  

In the following minimal example, the running state will be set when "Hello World!" is printed. 
[program:hello_world] 
command = /bin/sh -c  "sleep 10 && echo Hello World! && sleep 10" 
runningregex=Hello
stdout_logfile=./log/hello_world.log

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -935,7 +935,7 @@ class ServerOptions(Options):
         runningregex = get(section, 'runningregex', None)
         if runningregex:
             try:
-                runningregex = re.compile(runningregex)
+                runningregex = re.compile(r'.*'+ runningregex)
             except Exception as e:
                 raise ValueError(
                     f"program section {section} has invalid runningregex value. Error {e}")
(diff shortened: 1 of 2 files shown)


michaelhammann:
Appreciate all the effort you put into supervisord and the clear answer.
I fully understand that you do not want to add this feature to the core. However I believe this is a very useful functionality. For our use cases we take advantage of this for almost all programs. I will evaluate if we are running and maintaining our own fork or if we can provide a plugin accordingly.

PR #1449, opened 2021-07-16
feature: add optional parameter depends_on attribute to start dependent processes consecutively

michaelhammann:
This optional depends_on option is used to only start a process if the required process reached the RUNNING state.

In the following minimal example,  "Hello World!" followed by "done" is printed.
[program:hello_world]
command = /bin/sh -c "echo Hello World!"

[program:start_hello_world]
command =  /bin/sh -c "echo done"
depends_on=hello_world
autostart=true


supervisor/process.py
```diff
@@ -194,6 +195,33 @@ def spawn(self):
...
         """
+        # this section is ownly triggered when using autostart - 
+        # otherwise the dependencies are handled in the rpcinterface
+        # check if the process is dependent upon any other process and if so,
+        # make sure that one is in the RUNNING state
+        if self.config.depends_on is not None:
+            # keep track of RUNNING childs
+            running_childs = set()
+            # wait/loop until all childs are running
+            while set(self.config.depends_on.keys()) != running_childs:
```
alexsilva, line 206: The proposal looks good but code execution needs to be asynchronous. This code will stop the main supervisor loop.

michaelhammann: Thanks @alexsilva for the review. Stopping the main supervisor loop is not a good idea. I changed the implementation to use the rpcinterface also for spawning dependee processes when using autostart=true.  However, there might be better ways to use asyncronous calls, if you have any ideas on how to approve this any help is appreciated. Additionally there was a bug that spawn errors upon any dependee process were not properly detected. This should be fixed now.

michaelhammann: Quick update from myside. I was able to implement the depends_on flag without stopping the main supervisor loop. Currently testing it and then I will push the changes.

samssann:
Any update on when this could be added to a release?

michaelhammann:
I rebased to the current master now and will work on the missing unit tests, hopefully it will then  be part of any future release.

lfvjimisola:
@alexsilva Any chance that this will be or should we opt for an ad-hoc and less elegant solution?

michaelhammann:
If we want this feature to be added fast, one approach could be to try to implement it as a plugin, similar to what has been mentioned regarding a different feature request here: [](https://github.com/Supervisor/supervisor/pull/1446#issuecomment-1076707720) . Unfortunately I did not have time to have to have a closer look into creating this as a plugin yet. Happy to have a talk about it.

michaelhammann:
> @alexsilva Any chance that this will be or should we opt for an ad-hoc and less elegant solution?

If we want this feature to be added fast, one approach could be to try to implement it as a plugin, similar to what has been mentioned regarding a different feature request here: https://github.com/Supervisor/supervisor/pull/1446#issuecomment-1076707720  [.](https://github.com/Supervisor/supervisor/pull/1446#issuecomment-1076707720) Unfortunately I did not have time to have to have a closer look into creating this as a plugin yet. Happy to have a talk about it.

LorenzoRogai:
@michaelhammann is it possible to do the same for shutdown order?

PR #1481, opened 2021-12-21
Update http.py to support IPv6 connectivity

Veramkovich:
If IPv4 socket throws Exception then try to open socket with IPv6.

### DIFF
--- a/supervisor/http.py
+++ b/supervisor/http.py
@@ -524,9 +524,14 @@ class supervisor_af_inet_http_server(supervisor_http_server):
     def __init__(self, ip, port, logger_object):
         self.ip = ip
         self.port = port
-        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        self.prebind(sock, logger_object)
-        self.bind((ip, port))
+        try:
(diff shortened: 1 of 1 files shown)


supervisor/http.py
```diff
@@ -524,9 +524,14 @@ class supervisor_af_inet_http_server(supervisor_http_server):
...
         self.ip = ip
         self.port = port
-        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        self.prebind(sock, logger_object)
-        self.bind((ip, port))
+        try:
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            self.prebind(sock, logger_object)
+            self.bind((ip, port))
+        except:
```
mnaberez, line 531: I don't think we can do this.  If an IPv4 address was given and an error occurred, the error would be thrown away.  It would then try be tried as an IPv6 address, which would raise an unrelated error (probably `socket.error: [Errno 49] Can't assign requested address`).  This would make the original error very difficult to troubleshoot.  I think the code would have to determine if the address is IPv4 or IPv6 and then create the `AF_INET` or `AF_INET6` socket accordingly.

mnaberez (changes requested):

Issue #1101, opened 2018-05-16
kill leaves orphaned children


awronski:
Version: 3.3.1, Ubuntu 18.04

Configuration:
```
[program:xapi]
command=/usr/bin/npm run start
directory=/opt/nodeapps/xapi
user=nodejs
autostart=true
autorestart=true
startsecs=25
startretries=3
stopasgroup=true
killasgroup=true
stopwaitsecs=10
```

Process tree after start:
```
8307 nodejs     20   0 1098M 41928 25772 S  0.0  0.0  0:00.21 │  ├─ npm
8317 nodejs     20   0  4632   928   856 S  0.0  0.0  0:00.00 │  │  ├─ sh -c node -r dotenv/config dist/index.js
8318 nodejs     20   0  866M 38548 24768 S  0.0  0.0  0:00.16 │  │  │  └─ node -r dotenv/config dist/index.js
8323 nodejs     20   0  866M 38548 24768 S  0.0  0.0  0:00.00 │  │  │     ├─ node -r dotenv/config dist/index.js
8322 nodejs     20   0  866M 38548 24768 S  0.0  0.0  0:00.00 │  │  │     ├─ node -r dotenv/config dist/index.js
8321 nodejs     20   0  866M 38548 24768 S  0.0  0.0  0:00.00 │  │  │     ├─ node -r dotenv/config dist/index.js
8320 nodejs     20   0  866M 38548 24768 S  0.0  0.0  0:00.00 │  │  │     ├─ node -r dotenv/config dist/index.js
8319 nodejs     20   0  866M 38548 24768 S  0.0  0.0  0:00.00 │  │  │     └─ node -r dotenv/config dist/index.js
8316 nodejs     20   0 1098M 41928 25772 S  0.0  0.0  0:00.00 │  │  ├─ npm
8315 nodejs     20   0 1098M 41928 25772 S  0.0  0.0  0:00.00 │  │  ├─ npm
8314 nodejs     20   0 1098M 41928 25772 S  0.0  0.0  0:00.00 │  │  ├─ npm
8313 nodejs     20   0 1098M 41928 25772 S  0.0  0.0  0:00.00 │  │  ├─ npm
8312 nodejs     20   0 1098M 41928 25772 S  0.0  0.0  0:00.00 │  │  ├─ npm
8311 nodejs     20   0 1098M 41928 25772 S  0.0  0.0  0:00.00 │  │  ├─ npm
[...]

PR #1502, opened 2022-04-19
Scrape `ps` output to try and kill any grandchild processes when stopping a process

dat-h-ammann:
We also had problems with orphaned grandchild processes and that processes are still running even after stopping supervisor child processes. Here is a possible solution I want to share and I am happy for any feedback.

Use Case/ Context
Stopping a process via CLI or supervisor UI should stop the process and all it's child processes reliable. If the processes refuse to stop (e.g. not reacting to signals), they need to be forcefully stopped after some timeout. Additionally only direct child's of supervisor are properly stopped. Grandchild processes are not properly stopped and net be signaled as well.

Current Behavior
Supervisor handles different signals as stated here: http://supervisord.org/running.html#signals When a TERM, INT or QUIT signal is received, supervisor will trigger the shutdown behavior. This signals can also be sent to the process group, but not all processes started from a program are necessarily in the same process group, as mentioned here: https://github.com/Supervisor/supervisor/pull/199#issuecomment-132838482. Supervisor only knows about the direct child processes that it started but not any processes which any of the child processes spawned.

Possible Solution:
This stopping behavior still needs some improvements.
Stopping a process and all it's child processes reliable also needs a check if the processes are actually stopped and if the processes refuse to stop (e.g. not reacting to signals), they need to be forcefully stopped after some timeout via SIGKILL signal. There are a couple of things to consider:

think about a good timeout. Systemd uses about 2-3 mins, for my use-cases 60 seconds should be sufficient.
SIGKILL can be very intrusive. For example when processes are killed which are doing some operations on a database, this can cause loss of data.
There should be an additional config parameter such as disable_force_shutdown_behavior which would disable the mechanism of sending SIGKILL to the process and its childs.
We should define our usual behavior and allow users to alter this behavior. What's sensible is a default behavior -- I would say a round of SIGTERM to children, and after some time sending a SIGKILL to those who still alive is the ideal solution.

Note, however, that sometimes process can "hang" in the kernel (or as they call it D-state, see ps man page). It's rare in a healthy system, but still possible. For those situation, I would say we should report it to the user somehow.

Related to https://github.com/Supervisor/supervisor/issues/1101

### DIFF
--- a/supervisor/process.py
+++ b/supervisor/process.py
@@ -475,11 +475,17 @@ class Subprocess(object):
                 parent = psutil.Process(abs(pid))
                 children = parent.children(recursive=True)
                 for child in reversed(children):
-                    # kill all child processes with same signal as parent
-                    options.kill(child.pid, sig)
-                    # add to list but only if not disabled and supervisor instance provided
-                    if not self.config.disable_force_shutdown and sig is not signal.SIGKILL and supervisor is not None:
-                        supervisor.terminated_processes.append((child.pid, time.time()))
+                    try:
+                        # kill all child processes with same signal as parent
+                        options.kill(child.pid, sig)
+                        # add to list but only if not disabled
+                        if not self.config.disable_force_shutdown and sig is not signal.SIGKILL:
+                            supervisor.terminated_processes.append((child.pid, time.time()))
+                    except PermissionError:
+                        msg = ("No permission to signal child process with name '%s' and pid %i of parent process '%s' with pid %i. "
+                              % (child.name(), child.pid, parent.name(), parent.pid)
+                              + "Will continue with sending %s to parent process" % (sig.name))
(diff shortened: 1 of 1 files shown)


mrakh:
The ideal way to do this would be to designate `supervisord` as a "subreaper" so that it can inherit any orphaned descendant processes, then kill them. This functionality is supported by both Linux (with `prctl(PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0)`) and FreeBSD (with `procctl(P_PID, 0, PROC_REAP_ACQUIRE, NULL)`). You should be able to call those functions using `ctypes`.

Unfortunately, I can't find any kind of subreaper functionality for Solaris. But rather than executing `ps` and scraping its output, a more proper approach would be to walk the procfs to get all the descendant processes.

PR #1529, opened 2022-07-21
add maxbackoff to limit the interval between retries

jseparovic:


warrenspence:
I think this config option makes sense and I'd like to use it myself.

An issue with the PR is that you've set the default to 60 seconds which is effectively changing the existing behaviour for users currently providing a large `startretries` value.

I think if you set the default to 0 and then in `supervisor/process.py` only enforce the limit check if the `maxbackoff` is non 0 then the change would be safe to merge.

Issue #1540, opened 2022-09-10
Performance penalty for iteratively executing `os.close`


tyong920:
As per [Python official document on `os.closerange`](https://docs.python.org/3/library/os.html?highlight=os%20close#os.closerange)

`Close all file descriptors from fd_low (inclusive) to fd_high (exclusive), ignoring errors. Equivalent to (but much faster than):
`
```python
for fd in range(fd_low, fd_high):
    try:
        os.close(fd)
    except OSError:
        pass
```

It's better replace with it.
https://github.com/Supervisor/supervisor/blob/906d2adec9bd27d54cdff48bdd12f6cb42bf9546/supervisor/options.py#L1329-L1333

mnaberez:
Continued in #1541

PR #1541, opened 2022-09-10
Replace os.close with os.closerange

tyong920:
Issue #1540

### DIFF
--- a/supervisor/options.py
+++ b/supervisor/options.py
@@ -1326,11 +1326,7 @@ class ServerOptions(Options):
     def cleanup_fds(self):
         # try to close any leaked file descriptors (for reload)
         start = 5
-        for x in range(start, self.minfds):
-            try:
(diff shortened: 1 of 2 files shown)


mnaberez:
Description from #1540:

> As per [Python official document on `os.closerange`](https://docs.python.org/3/library/os.html?highlight=os%20close#os.closerange)
> 
> `Close all file descriptors from fd_low (inclusive) to fd_high (exclusive), ignoring errors. Equivalent to (but much faster than):
> `
> ```python
> for fd in range(fd_low, fd_high):
>    try:
>        os.close(fd)
>    except OSError:
>        pass
> ```
>
> It's better replace with it.
> https://github.com/Supervisor/supervisor/blob/906d2adec9bd27d54cdff48bdd12f6cb42bf9546/supervisor/options.py#L1329-L1333

PR #1546, opened 2022-09-25
Compression rotated logs feature

### LINKED ISSUE #322: Feature request: Have an option to compress log files after their rotation

winterheart:
This PR adds feature for compressing rotated backup files to reduce overall size of compressed files.
Added `*logfile_compression` options which controls how next backup file will be handled - copied or compressed. Supported formats are `copy` (default behavior), `gzip`, `bzip2` and `xz`.
`*logfile_backups` correctly handles both plain and compressed files in series, i.e. out.log.1, out.log.2.gz, out.log.3.bz2, out.log.4.xz files will be handled as one sequence of backup files.

Closes #322.


shouldsee:
what's going on with this PR?

joguy56:
This one would be a great one. Could avoid logrotate combinations.

supervisor/compressors.py
```diff
@@ -0,0 +1,59 @@
...
+
+class GzCompressor(Compressor):
+    def __init__(self, compressionLevel=9):
+        self.suffix = ".gz"
+        self.compressionLevel = compressionLevel
+
+    def compress(self, src, tgt):
+        with open(src, "rb") as f_src:
+            with gzip.open(tgt + self.suffix, "wb", compresslevel=self.compressionLevel) as f_tgt:
+                f_tgt.writelines(f_src)
```
mnaberez, line 37: Similar to [#619](https://github.com/Supervisor/supervisor/pull/619/files#r30990694), this kind of blocking code can't be used inside `supervisord`. It may take several seconds to compress the log file before control is returned to the event loop. During this time, `supervisord` won't be able to do any of its normal processing, e.g. it won't respond to commands from `supervisorctl`. 

Blocking in this way will also cause problems for the programs running under `supervisord`. If a program writes to its `stdout` file descriptor when `supervisord` is blocked compressing the log, `supervisord` will not be able to read the data from the pipe.  The pipe buffer can then become full.  If that happens, the program may hang in the write or the write may fail.

joguy56: Maybe this code could be reworked to make the compression part be asynchronous and not block the code ?

PR #1558, opened 2022-12-09
change serverurl option default value

wenjun93:
inet (TCP) server disabled by default, cli should use uds address by default for convinence


mnaberez:
> inet (TCP) server disabled by default, cli should use uds address by default for convinence

I think this is reasonable considering `inet_http_server` is disabled by default.   However, I do not want to change it because it has been the current default value for many years.  There are probably a lot of existing installations depending on the current default value and I don't want them to break unexpectedly when they upgrade Supervisor.

If you want the command `supervisorctl` (no arguments) to connect to a different `serverurl` than the default, you can create a config file in one of the [default locations](http://supervisord.org/configuration.html).  That file only needs a `[supervisorctl]`section.

wenjun93:
@mnaberez make sense

PR #1613, opened 2023-12-08
Redo "Parse env configuration in posix mode"

### LINKED ISSUE #328: Not possible to have quotes in environment variable settings

vereszol:
Fixes #328

This allows (escaped) quotes in the values as well as empty values.

This was done in pull request #329 but removed as it broke parsing empty quotes (#873) due to a bug in shlex (http://bugs.python.org/issue21999). This bug is fixed so posix mode can be used.

Brings back #329
Partially reverts #880

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,9 @@
   the ``[supervisorctl]`` section or ``[ctlplugin:x]`` sections to be in
   included files.  Patch by François Granade.
 
+- Parsing ``environment=`` has been improved to allow escaped quotes
+  inside quotes and quoted empty values. Patch by Stefan Friesel.
+
 4.2.5 (2022-12-23)
 ------------------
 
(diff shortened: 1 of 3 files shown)


vereszol:
https://bugs.python.org/issue21999 was fixed in Python 2.7 and 3.5 so this can be merged only after dropping support for Python 3.4

PR #1626, opened 2024-02-28
Add maxbackoff and backoffincrement options

cdelguercio:
* maxbackoff - the maximum number of seconds that supervisor will backoff before restart

* backoffincrement - the number of seconds to add to the backoff after each retry

based on https://github.com/Supervisor/supervisor/pull/1529

### DIFF
--- a/supervisor/process.py
+++ b/supervisor/process.py
@@ -168,7 +168,8 @@ class Subprocess(object):
         self.state = new_state
         if new_state == ProcessStates.BACKOFF:
             now = time.time()
-            self.backoff += 1
+            if self.backoff < self.config.maxbackoff or self.config.maxbackoff == 0:
+                self.backoff += self.config.backoffincrement
             self.delay = now + self.backoff
 
         event_class = self.event_map.get(new_state)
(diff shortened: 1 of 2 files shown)


cdelguercio:
If maxbackoff and backoffincrement are both set, adding backoffincrement to backoff may result in the remaining retries having a backoff value higher than maxbackoff.

Ex:
maxbackoff = 25
backoffincrement = 10

backoff will go from
1, 11, 21, 31, 31, 31, 31...

supervisor/process.py
```diff
@@ -168,7 +168,8 @@ def change_state(self, new_state, expected=True):
         self.state = new_state
         if new_state == ProcessStates.BACKOFF:
             now = time.time()
-            self.backoff += 1
+            if self.backoff < self.config.maxbackoff or self.config.maxbackoff == 0:
+                self.backoff += self.config.backoffincrement
```
warrenspence, line 172: @cdelguercio to prevent the issue you mentioned in https://github.com/Supervisor/supervisor/pull/1626#issuecomment-1969883652 you can use `min`:
```
self.backoff = min(self.backoff + self.config.backoffincrement, self.config.maxbackoff)
```

cdelguercio: I think you'd have to do something like:
```
self.backoff = self.config.maxbackoff == 0 ? self.backoff + self.config.backoffincrement : min(self.backoff + self.config.backoffincrement, self.config.maxbackoff);
```
The question is: what's preferred?

warrenspence: Yeah I see what you are saying, I missed the `self.config.maxbackoff == 0` case.

What you have suggested looks correct :+1:

Issue #1030, opened 2017-12-21
Supervisor does not detect any secondary group a user belongs to


wbytebier:
In our setup, we are running sssd, and none of the users used with Supervisor have entries in /etc/passwd or /etc/group.
In this case Supervisor only detects the primary group, and none of the secondary groups.

mnaberez:
`supervisord` doesn't dig around in any of those files itself.  [It calls](https://github.com/Supervisor/supervisor/blob/3.3.3/supervisor/options.py#L1261-L1306) Python's [`grp.getgrall()`](https://docs.python.org/2/library/grp.html#grp.getgrall) to get a list of groups and its members, finds all groups the user is a member of, then calls [`os.setgroups()`](https://docs.python.org/2/library/os.html#os.setgroups).  If the groups can't be found/set using these mechanisms, then `supervisord` can't be used.

lfrancke:
For anyone else stumbling across this: Python's `getgrall()` calls `getgrent()` under the hood which does not necessarily list groups from SSSD. You need to enable enumeration in SSSD for that to work (which also is not ideal due to its performance impact). RedHat has a Knowledge Base article for this: https://access.redhat.com/solutions/657713

gerases:
@lfrancke , this helped solve a long standing bug. Thank you!

PR #1631, opened 2024-03-12
ESRF's  addition for group IDs from sssd

fatboy92:
Gives child processes the effective Group IDs, even, when they are their definition is made in a sssd.
Solves https://github.com/Supervisor/supervisor/issues/1030

### DIFF
--- a/grouplist.py
+++ b/grouplist.py
@@ -1,0 +1,31 @@
+#!/usr/bin/python
+import pwd, os
+from ctypes import *
+from ctypes.util import find_library
+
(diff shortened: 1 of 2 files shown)


fatboy92:
I used the following command to list the groups, to which the user is a member:
grep '^Groups' /proc/${PID}/status
Groups: 125 125 505 1661 30634 32404 37454

mnaberez (changes requested):
Supervisor does not have any existing code using `ctypes` and we do not want to add any.  Please implement the feature using Python stdlib only.

Issue #1547, opened 2022-09-30
Release 4.2.5 or 4.3


piotrkochan:
Good morning, the 4.2.4 was released on Dec 30, 2021, that's a long time ago and some new functionalities have been merged since then. When are you going to release a new version? Is there anything stopping the new release?

mnaberez:
Supervisor [usually](https://github.com/Supervisor/supervisor/blob/4.2.4/CHANGES.rst) has one or two feature releases per year, along with whatever bug fix releases are needed between them.  There will be at least one new feature release of Supervisor before the end of the year.  New releases are posted to [PyPI](https://pypi.org/project/supervisor/) when they are available.

Issue #771, opened 2016-06-13
Read environment variables from file


standa4:
Hello,

is it possible read the environment variables from file or are you planning to add this feature?
It would be useful in case where I have multiple applications managed by supervisord and these applications share the same env vars. I don't want to edit multiple conf files. These variables can be stored in one file and read from it. 

Thank you.

mnaberez:
There is an `environment=` option in the [`[supervisord]`](http://supervisord.org/configuration.html#supervisord-section-values) section.  Put the variables in the `[supervisord]` section and they will be inherited by all subprocesses.

Choms:
Hi,

This is not an appropriate answer, you are not only not answering to @standa4's question, but you are also treating him in a condescending way, like if he was stupid enough as for not reading the documentation that clearly states how to use the `environment` line...

I'll exemplify his request with another example I think would be more helpful on depicting the issue... let's say I have a bash script with a conditional that checks some system parameters, and based on that sets one env with a value or a different one that I'll use in my program... it would be really helpful to be able of sourcing that file from the supervisord configuration file. (inb4 just create a bash wrapper that executes your program and call the wrapper from supervisord)

Now the **actual** question was "are you planning to add this feature?" - Can we please get a yes/no answer (I'm guessing it's a "no" based on the date of this issue and the other related issues that were closed too), and, if you feel generous enough, an argument for your decision?

Thanks

mnaberez:
> This is not an appropriate answer, you are not only not answering to @standa4's question, but you are also treating him in a condescending way, like if he was stupid enough as for not reading the documentation that clearly states how to use the environment line...

[The question](https://github.com/Supervisor/supervisor/issues/771#issue-159957429) asked for a single place to define environment variables, citing a desire to not modify multiple config files.  Putting the variables in `[supervisord]` section of the root config file seems to meet those requirements.   There are two places that `environment=` can be specified in the config file(s).  One is any `[program:x]` section where the variables will be applied only to that program.  The other is in the `[supervisord]` section, where the variables will inherited by all programs.  My response was intended to be helpful, not condescending.  There's a lot of documentation, and it's easy to miss details like this.

> Now the actual question was "are you planning to add this feature?" - Can we please get a yes/no answer (I'm guessing it's a "no" based on the date of this issue and the other related issues that were closed too), and, if you feel generous enough, an argument for your decision?

There are not currently plans to add a separate type of config file just for environment variables.  Since this has already been discussed in other issues, please refer to those issues.

[...]

PR #1632, opened 2024-03-14
add envfile option

sailist:
As #370  #771 #1547 mentioned, this is an useful feature. Actually it's easy to implement.


mnaberez:
Duplicate of #934

Issue #1391, opened 2020-11-27
improve secondary group lookup performance with LDAP/NIS


zegab1:
The current implementation of supervisord does a serial enumeration of all existing groups on the host to determine secondary group membership of the effective user of the managed service to be started. This is could take several minutes if group resolution is done over the network and if there are a few thousand groups are available.

zegab1:
In Python 2.7+ we have the os.initgroups() call, which can do an NSS specific call to initialize group membership of a user, thus could be used to speed up the launch of the managed process. I have used the following patch successfully in such an environment:

`--- options.py.orig	2020-11-27 13:45:39.772461100 +0100
+++ options.py	2020-11-27 13:53:49.644546000 +0100
@@ -1378,15 +1378,27 @@
         gid = pwrec[3]
         if hasattr(os, 'setgroups'):
             user = pwrec[0]
-            groups = [grprec[2] for grprec in grp.getgrall() if user in
-                      grprec[3]]
 
-            # always put our primary gid first in this list, otherwise we can
-            # lose group info since sometimes the first group in the setgroups
-            # list gets overwritten on the subsequent setgid call (at least on
-            # freebsd 9 with python 2.7 - this will be safe though for all unix
-            # /python version combos)
-            groups.insert(0, gid)
+            # add support for os.initgroups() in Python 2.7+
+            # after calling os.initgroups(), we retrieve the current group list
+            # with os.getgroups() which is way faster than the sequential scanning
+            # of grp.getgrall() if groups are resolved over LDAP/NIS
+            if hasattr(os, 'initgroups'):
+               try:
+                   os.initgroups(user,gid)
+                   groups = os.getgroups()
+               except OSError:
+                   return 'Could not init groups of effective user'
+            else:
+               groups = [grprec[2] for grprec in grp.getgrall() if user in
+                       grprec[3]]
+               # always put our primary gid first in this list, otherwise we can
+               # lose group info since sometimes the first group in the setgroups
+               # list gets overwritten on the subsequent setgid call (at least on
+               # freebsd 9 with python 2.7 - this will be safe though for all unix
+               # /python version combos)
+               groups.insert(0, gid)
+
             try:
                 os.setgroups(groups)
             except OSError:
`

mnaberez:
> initgroups.patch.gz

```diff 
--- options.py.orig	2020-11-27 13:45:39.772461100 +0100
+++ options.py	2020-11-27 13:53:49.644546000 +0100
[...]

PR #1634, opened 2024-03-18
ESRF addition to creating processes with all GIDs.

fatboy92:
At ESRF we need data acquisition process to create files with read and write permissions for all groups of the user.

Hopefully this change fulfills your request to avoid loading libc.


mnaberez:
Related: https://github.com/Supervisor/supervisor/issues/1391#issuecomment-734955455 is another request to change how groups are assigned.

PR #1643, opened 2024-04-30
Bugfix: Updated expansions before calling expand() function for envir…

Tanvi840:
…onment variables

Bugfix: Variables defined in environment= in [program:x] section cannot be used in %(ENV_X)s expansions since expansions are updated after calling the expand function.


Tanvi840:
Incorrect fix

PR #1651, opened 2024-07-18
Add missing channel field to process log event docs

liamadamson:
Process log events (from both stdout and stderr) have channel
information in their payload. The docs have been updated to reflect
this.


mnaberez:
Thank you!

PR #1668, opened 2024-12-12
Fix tests on Python 3.13

cjwatson:
Python 3.13 added a [`__firstlineno__`](https://docs.python.org/3/reference/datamodel.html#type.__firstlineno__) attribute to classes whose value is an int, so the approach taken by
`SubprocessTests.test_getProcessStateDescription` to test only actual states no longer works.  Exclude all attributes starting with `__` instead.


mnaberez:
Thanks!

PR #1705, opened 2025-12-07
Do not swallow error when importing plugins

SuperSandro2000:
I had an error in knot-resolver v6 when importing its plugin which before looked like:

```
Error: knot_resolver.controller.supervisord.plugin.patch_logger:inject cannot be resolved within [rpcinterface:patch_logger]
```

which was not very actionable or clear for me.

Now the error is:

```
Error: knot_resolver.controller.supervisord.plugin.patch_logger:inject cannot be resolved within [rpcinterface:patch_logger]: No module named 'knot_resolver'
```

### DIFF
--- a/supervisor/tests/test_options.py
+++ b/supervisor/tests/test_options.py
@@ -2644,7 +2644,8 @@ class ServerOptionsTests(unittest.TestCase):
         except ValueError as exc:
             self.assertEqual(exc.args[0],
                 'thisishopefullynotanimportablepackage:nonexistent cannot be '
-                'resolved within [eventlistener:cat]')
+                'resolved within [eventlistener:cat]: No module named '
+                '\'thisishopefullynotanimportablepackage\'')
 
     def test_event_listener_pool_result_handler_unimportable_AttributeError(self):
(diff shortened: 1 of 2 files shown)


supervisor/tests/test_options.py
```diff
@@ -2644,7 +2644,8 @@ def test_event_listener_pool_result_handler_unimportable_ImportError(self):
         except ValueError as exc:
             self.assertEqual(exc.args[0],
                 'thisishopefullynotanimportablepackage:nonexistent cannot be '
-                'resolved within [eventlistener:cat]')
+                'resolved within [eventlistener:cat]: No module named '
+                '\'thisishopefullynotanimportablepackage\'')
```
SuperSandro2000, line 2648: I am not sure how to handle the quoting issues here.

mnaberez: Fixed in 83eeeaf3107b19329ba556608003f189bbcb590f. Thanks for the patch!

SuperSandro2000: Thanks 🎉

PR #1709, opened 2026-01-28
Indefinite, fast restarts when 'startretries' = -1

tellscode:
Hi, first many thanks for developing supervisor!

This pull request updates the behavior and documentation of the startretries configuration option to support infinite retries when set to -1. Also it is skipping the otherwise always increasing delays.

When developing in an environment with hot-reload, I found myself sometimes being restricted by long restart times. This change solved this for me. I'd be very happy if that change would make it in the official repo, but I can also understand if this shouldn't be a part of supervisor.

Best regards,
Tell

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -752,7 +752,8 @@ where specified.
 
   The number of serial failure attempts that :program:`supervisord`
   will allow when attempting to start the program before giving up and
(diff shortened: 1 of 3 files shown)


jacksgt (approved):
Very nice! I was recently looking for a similar feature!

PR #1717, opened 2026-05-26
supervisord: skip idle reaping without child pids

Komzpa:
## Summary

`Supervisor.runforever()` currently calls `reap()` on every loop iteration after polling and process state transitions. When there are no tracked child PIDs, this still asks `waitpid()` for child status on every idle tick.

This change skips the periodic reap unless Supervisor has tracked child PIDs. The main loop reads the pending signal once and passes it to `handle_signal()`, so `SIGCHLD` is still logged through the existing signal path without an extra idle `waitpid()` call.

## Validation

- `python3 -m pytest supervisor/tests/test_supervisord.py -q`
- `END_TO_END=1 python3 -m pytest supervisor/tests/test_end_to_end.py::EndToEndTests::test_issue_1170a -q`
- `python3 -m pytest -q`

### DIFF
--- a/supervisor/supervisord.py
+++ b/supervisor/supervisord.py
@@ -45,6 +45,8 @@ from supervisor import events
 from supervisor.states import SupervisorStates
 from supervisor.states import getProcessStateDescription
 
+_NO_SIGNAL = object()
+
 class Supervisor:
     stopping = False # set after we detect that we are handling a stop request
     lastshutdownreport = 0 # throttle for delayed process error reports at stop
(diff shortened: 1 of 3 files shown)


mnaberez:
I find the patch much more confusing than the original code (particularly the `handle_signal(self, sig=_NO_SIGNAL)` bit) but moreover, I don't see this as a problem that needs to be solved.

`supervisord`'s job is to keep processes running.  Its normal operating state is that it has a child process running and it needs to call `waitpid` each time around the loop.  I see this patch as a micro-optimization: It adds a bunch of conditional logic to avoid calling `waitpid` in the loop when `supervisord` has no children.  As soon as a process starts, `waitpid` is going to get called anyhow.

I don't think this is something we should care about enough to make the code more difficult to understand.

